PR #3, opened 2025-04-14
[CLD-127]: ci: setup go releaser

graham-chainlink:
- Add push tag workflow
- setup goreleaser config
- npm commands to running changeset release

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-127

### DIFF
--- a/.github/workflows/push-tag-release.yml
+++ b/.github/workflows/push-tag-release.yml
@@ -1,0 +1,41 @@
+name: push-tag-release
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  cicd-publish-release:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: write
+      actions: read
+    steps:
+      - name: cicd-publish-release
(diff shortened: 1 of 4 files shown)


.github/workflows/push-tag-release.yml
```diff
@@ -0,0 +1,66 @@
...
+
+  ci-test:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+      actions: read
+    steps:
+      - name: ci-test
+        uses: smartcontractkit/.github/actions/ci-test-go@eeb76b5870e3c17856d5a60fd064a053c023b5f5 # ci-test-go@1.0.0
```
jkongie, line 32: Do we actually need this to run on release?

Since it's not a `needs` on the publish, they are just running and are not a blocker for a release

graham-chainlink: oh nice pick up, i copied this from MCMS, but i think you are right, the changesets raise a PR before pushing a tag, that PR will have all the required ci checks so we dont have to duplicate here again. 

Removing it.

PR #8, opened 2025-04-15
[CLD-160]: ci: setup changeset workflow

graham-chainlink:
Add push main workflow to trigger changeset bot and version PR

Also set the version in package.json to 0.0.1 to start with.

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-160

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module chainlink-deployments-framework
+module github.com/smartcontractkit/chainlink-deployments-framework
 
 go 1.24.1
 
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module chainlink-deployments-framework
+module github.com/smartcontractkit/chainlink-deployments-framework
```
graham-chainlink, line 1: This was incorrect before, module should have the full github name

PR #29, opened 2025-04-22
AddressBook to DataStore adapter

akuzni2:
Adds an adapter to convert between AddressBook and Datastore.

Some components still use AddressBook instead of the newer DataStore. This is helpful to temporarily allow components to switch back from DataStore to AddressBook for backward compatibility so that components using either AddressBook or DataStore can function while waiting for the common codebase to fully upgrade from AddressBook to DataStore.

### DIFF
--- a/deployment/address_book_adapter.go
+++ b/deployment/address_book_adapter.go
@@ -1,0 +1,72 @@
+package deployment
+
+import (
+	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
+)
+
+// AddressBookToDataStore converts an AddressBook to a DataStore
+// AddressBook is deprecated and will be removed in the future. You can use this function to migrate or interact with legacy code that uses AddressBook.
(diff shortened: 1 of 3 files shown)


akuzni2:
I am using this pattern in my changesets currently as MCMS changeset, deploy link token changeset are not using datastore yet

deployment/address_book_adapter.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	for chainSelector, chainAddresses := range addresses {
+		for address, tv := range chainAddresses {
+			addressRef := datastore.AddressRef{
+				Address:       address,
+				ChainSelector: chainSelector,
+				Labels:        datastore.LabelSet(tv.Labels),
+				Type:          datastore.ContractType(tv.Type),
+				Version:       &tv.Version,
```
giogam, line 34: Just a heads-up: if you don’t set a qualifier for the address ref, any duplicate combination of chain selector, contract type, and version will generate the same key and cause a clash. So if your addressbook has multiple records with the same contract, version, and chain, add will throw an error.

PR #30, opened 2025-04-24
Data store/add address filter

akuzni2:
Add an address filter. Usecase I have for this is to quickly check if the contract address I have is the expected one
```
records := store.Filter(
			AddressRefByAddress("0x1234"),
			AddressRefByChainSelector(1),
			AddressRefByType(ContractType("MyType")),
			AddressRefByVersion("0.5.0"),
		)

if len(records) == 1 .... 

```

It probably should have it's own dedicated function in the datastore to get a single expected record.

### DIFF
--- a/.changeset/yummy-clouds-argue.md
+++ b/.changeset/yummy-clouds-argue.md
@@ -1,0 +1,5 @@
+---
+"chainlink-deployments-framework": patch
+---
+
+Add an address filter to datastore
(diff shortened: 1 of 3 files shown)


jkongie:
This seems like a reasonable filter

@giogam This looks good to me, but I'll defer to you in case there was a reason why this was not included in the first place.

PR #38, opened 2025-04-26
fix(labels):init when add is nil, change LabelSet to pointer

ajaskolski:
This pull request refactors the `LabelSet` type in `datastore/label_set.go` to improve its functionality and safety by using pointer receivers and adding test cases for edge scenarios. The most important changes include switching `LabelSet` methods to use pointer receivers, adding safeguards for uninitialized `LabelSet` instances, and enhancing test coverage.

### Refactoring `LabelSet` methods:

* Updated all methods in `LabelSet` to use pointer receivers instead of value receivers to ensure modifications affect the original instance. (`datastore/label_set.go`, [[1]](diffhunk://#diff-1d4afca2b7bb8da9ebeeddbe1df26822f654e71505745ee7d3be234e3d16f8aaL28-R41) [[2]](diffhunk://#diff-1d4afca2b7bb8da9ebeeddbe1df26822f654e71505745ee7d3be234e3d16f8aaL47-R50) [[3]](diffhunk://#diff-1d4afca2b7bb8da9ebeeddbe1df26822f654e71505745ee7d3be234e3d16f8aaL58-R61) [[4]](diffhunk://#diff-1d4afca2b7bb8da9ebeeddbe1df26822f654e71505745ee7d3be234e3d16f8aaL73-R95) [[5]](diffhunk://#diff-1d4afca2b7bb8da9ebeeddbe1df26822f654e71505745ee7d3be234e3d16f8aaL97-R104)

* Added a safeguard in the `Add` method to initialize the `elements` map if it is `nil`, preventing runtime errors when modifying an uninitialized `LabelSet`. (`datastore/label_set.go`, [datastore/label_set.goL28-R41](diffhunk://#diff-1d4afca2b7bb8da9ebeeddbe1df26822f654e71505745ee7d3be234e3d16f8aaL28-R41))

### Enhancing test coverage:

* Added a test case to verify the behavior of adding a label to an uninitialized `LabelSet`. (`datastore/label_set_test.go`, [datastore/label_set_test.goR63-R72](diffhunk://#diff-799b56dc5bc27b050cbfd84aa25abb575627f45e22e691c96ff3fc24ff38d661R63-R72))

* Expanded the `Clone` method tests to include scenarios for cloning both non-empty and empty `LabelSet` instances, ensuring correctness and independence of the cloned object. (`datastore/label_set_test.go`, [[1]](diffhunk://#diff-799b56dc5bc27b050cbfd84aa25abb575627f45e22e691c96ff3fc24ff38d661R301) [[2]](diffhunk://#diff-799b56dc5bc27b050cbfd84aa25abb575627f45e22e691c96ff3fc24ff38d661L299-R321)

* Updated the `MarshalJSON` test to use a pointer reference for `LabelSet` to align with the updated method signature. (`datastore/label_set_test.go`, [datastore/label_set_test.goL333-R353](diffhunk://#diff-799b56dc5bc27b050cbfd84aa25abb575627f45e22e691c96ff3fc24ff38d661L333-R353))


datastore/address_ref.go
```diff
@@ -33,7 +33,7 @@ type AddressRef struct {
 	// ChainSelector is the chain-selector of the chain where the contract is deployed.
 	ChainSelector uint64 `json:"chainSelector"`
 	// Labels are the labels associated with the contract.
-	Labels LabelSet `json:"labels"`
+	Labels *LabelSet `json:"labels"`
```
jkongie, line 36: Actually I would leave this as a value instead of a pointer. I would rather always have the labels instead of having to deal with nils all the way down

ajaskolski: you are right seems easier there, updated.

datastore/label_set.go
```diff
@@ -94,7 +101,7 @@ func (s LabelSet) Clone() LabelSet {
 // MarshalJSON marshals the LabelSet as a JSON array of strings.
 //
 // Implements the json.Marshaler interface.
-func (s LabelSet) MarshalJSON() ([]byte, error) {
+func (s *LabelSet) MarshalJSON() ([]byte, error) {
```
jkongie, line 104: The `MarshalJSON` method should be on the value receiver not the pointer receiver. 

This is because "Pointer values encode as the value pointed to."

datastore/label_set.go
```diff
@@ -70,22 +73,26 @@ func (s LabelSet) List() []string {
...
+func (s *LabelSet) IsEmpty() bool {
 	return s.Length() == 0
 }
 
 // Clone creates a copy of the LabelSet.
-func (s LabelSet) Clone() LabelSet {
+func (s *LabelSet) Clone() LabelSet {
+	if len(s.elements) == 0 {
+		return NewLabelSet()
+	}
```
jkongie, line 94: by adding this method, we are not truly cloning the struct, since we modify the underlying map. I think it would be better to remove this

datastore/label_set.go
```diff
@@ -25,17 +25,20 @@ func NewLabelSet(labels ...string) LabelSet {
...
+func (s *LabelSet) Add(label string) {
+	if s.elements == nil {
+		s.elements = make(map[string]struct{})
+	}
 	s.elements[label] = struct{}{}
 }
 
 // Remove deletes a label from the set, if it exists.
-func (s LabelSet) Remove(label string) {
+func (s *LabelSet) Remove(label string) {
```
jkongie, line 36: This is good. I didn't take into account that when this was using a type alias, the aliased map is already a pointer, so using a value receiver worked. 

Thanks for fixing this

datastore/label_set.go
```diff
@@ -25,17 +25,20 @@ func NewLabelSet(labels ...string) LabelSet {
 }
 
 // Add inserts a label into the set.
-func (s LabelSet) Add(label string) {
+func (s *LabelSet) Add(label string) {
+	if s.elements == nil {
+		s.elements = make(map[string]struct{})
+	}
```
jkongie, line 31: 👍

datastore/label_set.go
ajaskolski: oh interesting I didn't know that. Thank you for clarification on that !

datastore/label_set.go
ajaskolski: ok I see `maps.Clone` handles nil/empty inside correctly so no need for that

jkongie:
Any reason why the pointer receiver was removed from `Clone`? I believe it should be OK to use a pointer there

jkongie (approved):

PR #39, opened 2025-04-29
feat: change operations retry logic to be opt-in

jkongie:
BREAKING CHANGE: The retry logic for operations is now opt-in. To enable the retry, the caller must now provide the `WithRetry` option to the `ExecuteOperation`. This change is intended to provide greater control over the retry behavior and to avoid unintended retries in certain scenarios.

### DIFF
--- a/.changeset/tangy-plums-draw.md
+++ b/.changeset/tangy-plums-draw.md
@@ -1,0 +1,5 @@
+---
+"chainlink-deployments-framework": patch
+---
+
+BREAKING: Operations retry logic is now opt in. Use the `WithRetry` method in your `ExecuteOperation` call to enable retries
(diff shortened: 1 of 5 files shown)


ajaskolski (approved):

.changeset/tangy-plums-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+"chainlink-deployments-framework": patch
```
graham-chainlink, line 2: If this is breaking, should this be a major?

graham-chainlink (approved):
looks good, we should update the docs too when this goes in

.changeset/tangy-plums-draw.md
jkongie: any 0.X.X release can have breaking changes. It is only until we go to 1.X.X that we become stable

graham-chainlink: gotcha! Just checking!

PR #43, opened 2025-04-29
feat: adds debug logs to ethclient direct calls

giogam:
This PR extends the MultiClient implementation with additional method overrides to help surface errors originating from the underlying ethclient. The goal is to improve visibility into the flakiness observed when running changesets through CI and GAP, and to help identify the specific call causing the issue. 

The additional overrides cover all method calls that can be invoked as part of a `transact` call, as seen [here](https://github.com/ethereum/go-ethereum/blob/612c9e0f4ae483f180ef96ace86effec6dc9bb0e/accounts/abi/bind/base.go#L390).

### DIFF
--- a/deployment/multiclient.go
+++ b/deployment/multiclient.go
@@ -305,6 +305,26 @@ func (mc *MultiClient) BalanceAt(ctx context.Context, account common.Address, bl
 	return balance, err
 }
 
+func (mc *MultiClient) FilterLogs(ctx context.Context, q ethereum.FilterQuery) ([]types.Log, error) {
+	var logs []types.Log
+	err := mc.callWithDebug("FilterLogs", func(client *ethclient.Client) error {
+		var err error
+		logs, err = client.FilterLogs(ctx, q)
+		return err
+	})
+	return logs, err
+}
+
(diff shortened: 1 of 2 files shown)


deployment/multiclient.go
```diff
@@ -231,3 +231,85 @@ func (mc *MultiClient) dialWithRetry(rpc RPC, lggr logger.Logger) (*ethclient.Cl
...
+		return err
+	})
+	return balance, err
+}
+
+func (mc *MultiClient) callWithDebug(callName string, call func(*ethclient.Client) error) error {
+	var err error
+	err = call(mc.Client)
+	if err != nil {
+		mc.lggr.Debugf("error '%s' for call %s for chain %s", MaybeDataErr(err), callName, mc.chainName)
```
chainchad, line 312: Should we change this to an error log (instead of debug) and add a debug log when this function is called to show what the `callName` is?

jkongie (commented):
We should probably replicate the delegate calls in this implementation of a retry client. This is used in rddtool-ccip

https://github.com/smartcontractkit/rddtool-ccip/blob/master/internal/rpclib/ethclient_gen.go

deployment/multiclient.go
giogam: Yeah, good call. I was thinking we could just log in case of an error, but since these calls aren’t retried, they shouldn’t flood the logs too much.

jkongie (approved):

PR #59, opened 2025-05-06
[CLD-210]: feat(mutliclient): better log & expose dial config

graham-chainlink:
- better logging with trace id, in logs where there are multiple calls with the same chain name, we are unable to tell if the retry attempt logs are from the same call or different, having a traceID will make this more obvious and help with debugging.
- expose dial attempts and delay as external config so it can be configured.

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-210

### DIFF
--- a/deployment/multiclient.go
+++ b/deployment/multiclient.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/google/uuid"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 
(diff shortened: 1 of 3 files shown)


deployment/multiclient.go
```diff
@@ -30,14 +31,18 @@ const (
 )
 
 type RetryConfig struct {
-	Attempts uint
-	Delay    time.Duration
+	Attempts     uint
+	Delay        time.Duration
+	DialAttempts uint
```
DimitriosNaikopoulos, line 36: Do we need to have different configs for RPC calls and dialing endpoints? I can see that the current configuration was like that but is there a reason that we may need that in the future since currently both have the same values? 🤔

graham-chainlink: Good question, i think it is useful, i am not the original author of the code but my believe is that dial and calls are for different purpose, thus can require different retry characteristic (even though they are the same now) 

Dial is for establishing connection, usually we want lower max attempts and shorter delay since if we can’t connect quickly, the service is likely down or misconfigured.

Where as Call is for actual operations, so we want to be more generous and have higher max attempts and longer delay (as it is a heavier operation)

Maybe @giogam knows more

DimitriosNaikopoulos (approved):

PR #69, opened 2025-05-08
add solana program close

tt-cll:


deployment/solana_chain.go
```diff
@@ -91,6 +91,35 @@ func GetSolanaProgramBytes() map[string]int {
...
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	// Run the command
+	if err := cmd.Run(); err != nil {
+		logger.Errorw("Error closing buffers",
+			"error", err,
+			"stdout", stdout.String(),
+			"stderr", stderr.String())
+	}
```
graham-chainlink, line 117: should there be a return or else statement around here? if there is an error, we still print "Closed buffers'

tt-cll: yes good call

graham-chainlink (approved):

PR #72, opened 2025-05-08
add custom JSON marshaling for DefaultMetadata

akuzni2:
Added a custom `MarshalJSON` method to `DefaultMetadata`. This ensures that if the `Data` field contains JSON, it is not double-encoded, while non-JSON data is marshaled as-is.


.changeset/famous-doors-ask.md
```diff
@@ -0,0 +1,5 @@
+---
+"chainlink-deployments-framework": patch
+---
+
+marshall json without double-encoding
```
graham-chainlink, line 5:
```suggestion
marshal json without double-encoding
```

graham-chainlink (commented):
You got a lint error

datastore/default_metadata.go
```diff
@@ -1,9 +1,31 @@
...
 	Data string `json:"data"`
 }
 
 // DefaultMetadata implements the Cloneable interface
-func (d DefaultMetadata) Clone() DefaultMetadata { return d }
+func (dm DefaultMetadata) Clone() DefaultMetadata { return dm }
+
+// MarshalJSON can handle both JSON and non-JSON data in the Data field.
+// Without custom marshaling, the Data field would be double-encoded if it contains JSON.
+func (dm DefaultMetadata) MarshalJSON() ([]byte, error) {
```
graham-chainlink, line 15: Then unmarshaling would not work right since the Json can't be converted back to a string unless we implement Unmarshaljson too? Let me check with @giogam to understand this a bit more.

akuzni2: That's a good point - we'd need to implement that

giogam: Hey @akuzni2! Can you give us more context on why this is needed?  Ideally the `Data` field is intended to store a string JSON representation of a domain custom metadata type.

akuzni2: Without correct serialization
```
       {
          "contractMetadataStore": {
            "records": [
              {
                "address": "0x1234",
                "chainSelector": 3478487238524512106,
                "metadata": {
                  "data": "{\"content\":{\"metadata\":{\"deployBlock\":150906864},\"view\":{\"field1\":[\"0x1234567890abcdef\"],\"field2\":{\"key1\":\"value1\"}}}}"
                }
              }
            ]
          },
        }
```     

With this custom serializer

```
       {
          "contractMetadataStore": {
            "records": [
              {
                "address": "0x1234",
                "chainSelector": 3478487238524512106,
                "metadata": {
                  "data": {
                    "content": {
                      "metadata": {
                        "deployBlock": 150906864
                      },
                      "view": {
                        "field1": [
                          "0x1234567890abcdef"
                        ],
                        "field2": {
                          "key1": "value1"
                        }
                      }
                    }
                  }
                }
              }
            ]
          },
        }   
```

Isn't the latter the intended format? 

The latter would be the proper way to store JSON. It's breaking right now because ToDefault does a cast string(data)
https://github.com/smartcontractkit/chainlink-deployments-framework/blob/6968d63e8c6db17afa95cdd45748392d5c15a5af/datastore/transform.go#L49

if the `Data` field of DefaultMetadata is really intended to be JSON - we should change this to Go's json.RawMessage which will handle the double encoding for us. 

So theres 2 choices:
1) custom serialization (implemented in PR)
2) break the data-type and hope no one is using it directly and go and update it everywhere

graham-chainlink:
Thanks Alex, I will leave this to @giogam as he has a lot more context on  the design of datastore.

giogam:
Closing this, as we ended up taking a different approach.
Refer to the following PRs for more details:

- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/124
- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/141
- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/149
- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/118

PR #84, opened 2025-05-15
feat(multiclient): adds timeouts to retryWithBackups and dialWithRetry

giogam:
This PR enhances the `MultiClient` implementation by adding timeout support and improving test coverage.

Timeout Handling:

- Added `Timeout` and `DialTimeout` to `RetryConfig`, with defaults for RPC and dial operations.
- Updated `retryWithBackups` and `dialWithRetry` to use `context.WithTimeout`, enforcing timeouts for RPC calls and connections.

Context Propagation:

- All MultiClient methods (SendTransaction, CallContract, etc.) now pass context.Context through to support cancellation and timeouts consistently.

Test Improvements:

- Added `TestMultiClient_dialWithRetry` to cover dial timeouts and failure scenarios.
- Expanded `TestMultiClient_retryWithBackups` to simulate timeouts and long-running operations using context.


graham-chainlink (approved):

deployment/multiclient_test.go
```diff
@@ -49,26 +50,101 @@ func TestMultiClient(t *testing.T) {
...
+				lggr:        lggr,
+			}
+
+			_, err := mc.dialWithRetry(RPC{
+				Name:               "test-rpc",
+				WSURL:              tt.URL,
+				PreferredURLScheme: URLSchemePreferenceWS,
+			}, lggr)
+
+			require.Error(t, err)
```
graham-chainlink, line 102: i think checking what error here is also important so we can distinguish a deadline error or an bad url error

graham-chainlink:
So we are exposing context for each public method in this PR, which means user can pass down their own context with timeout, but then inside each method we override it with the RetryConfig timeout value, should we honour their timeout too or not? 

I would expect that if a method accepts context , it should uses it like `DialContext` in postgres.

A thought is that if we really want to enforce our own timeout, maybe we could set out own timeout if there are no existing timeout set in the user pass in context? Not super sure, lets see what everyone else thinks.

```go
func WithTimeoutIfNone(parent context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
    // Check if the parent context already has a deadline
    if _, hasDeadline := parent.Deadline(); hasDeadline {
        // Return the original context and a no-op cancel function
        return parent, func() {}
    }

    // No deadline: apply timeout
    return context.WithTimeout(parent, timeout)
}

// usage

ctx, cancel := WithTimeoutIfNone(ctx, DEFAULT_TIMEOUT)
```

Also with WaitMined, then now there is some inconsistency, a new user could set timeout on the RetryConfig but then doent know that it doesnt affect WaitMined but it affect the rest, a surprise?

giogam:
> So we are exposing context for each public method in this PR, which means user can pass down their own context with timeout, but then inside each method we override it with the RetryConfig timeout value, should we honour their timeout too or not?

Good question. I looked into the `ethclient` implementation and contract bindings, and it does appear possible to inject a context from a `changeset`. The contract bindings accept context via `TransactOpts`, see [here](https://github.com/smartcontractkit/chainlink-evm/blob/787b834d1520a341adaf07e7165a1f82d682b807/gethwrappers/shared/generated/link_token/link_token.go#L42) and [here](https://github.com/smartcontractkit/chainlink-evm/blob/787b834d1520a341adaf07e7165a1f82d682b807/gethwrappers/shared/generated/link_token/link_token.go#L236).  
 The `ethclient` implementation handles this in a similar way to what you suggested, for example, [here](https://github.com/ethereum/go-ethereum/blob/892a661ee2e7bcafcfd4eb981abf0ba697d57a5c/accounts/abi/bind/v2/base.go#L289). I’d say it’s unlikely someone is doing this, but it’s definitely possible. I’m happy to update the code with your suggestion, it’s a small change anyway.

> Also with WaitMined, then now there is some inconsistency, a new user could set timeout on the RetryConfig but then doent know that it doesnt affect WaitMined but it affect the rest, a surprise?

For `WaitMined`, I think it makes sense to have a separate, and potentially longer, timeout, since transactions can sometimes take a while to get mined. We could improve the code to make this configurable, but I’d still keep it as a separate setting from the client retry config.

graham-chainlink:
> For `WaitMined`, I think it makes sense to have a separate, and potentially longer, timeout, since transactions can sometimes take a while to get mined. We could improve the code to make this configurable, but I’d still keep it as a separate setting from the client retry config.

Sounds good! lets make sure we comment it so users know as well

deployment/multiclient.go
```diff
@@ -270,16 +276,19 @@ func (mc *MultiClient) FilterLogs(ctx context.Context, q ethereum.FilterQuery) (
...
+		sub, err = client.SubscribeFilterLogs(ct, q, ch)
 
 		return err
 	})
 
 	return sub, err
 }
 
+// WaitMined waits for a transaction to be mined and returns the receipt.
+// A timeout for this operation must be set in the context.
```
graham-chainlink, line 290: "must" is a strong word, is it true that we must set the timeout ourselves?

graham-chainlink (approved):

graham-chainlink (approved):

PR #85, opened 2025-05-16
[CLD-208]: fix(chain): refactor solana DeployProgram to accept ProgramBytes

graham-chainlink:
These constants are product specific and should not be coupled with the DeployProgram function. This refactors the method to accept programBytes instead and I will be moving the program bytes map back into Chainlink repo

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-208


jkongie (approved):

tt-cll:
Can we please not merge things like this without checking with the product team?

PR #88, opened 2025-05-19
rpc health check

bytesizedroll:
Dialing just establishes a TCP connection to the server, while an RPC call actually verifies the server is a functioning node by making a real request. This ticket tracks adding a quick RPC call to client creation to ensure the client we return to the user is usable.

### DIFF
--- a/deployment/multiclient.go
+++ b/deployment/multiclient.go
@@ -30,6 +30,9 @@ const (
 	RPCDefaultDialRetryAttempts = 1
 	RPCDefaultDialRetryDelay    = 1000 * time.Millisecond
 	RPCDefaultDialTimeout       = 10 * time.Second
+
+	// Default timeout for health checks
+	RPCDefaultHealthCheckTimeout = 2 * time.Second
 )
 
 type RetryConfig struct {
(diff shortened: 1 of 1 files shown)


deployment/multiclient.go
```diff
@@ -87,6 +103,11 @@ func NewMultiClient(lggr logger.Logger, rpcsCfg RPCConfig, opts ...func(client *
 			lggr.Warnf("failed to dial client %d for RPC '%s' trying with the next one: %v", i, rpc.Name, err)
 			continue
 		}
+		if err := mc.rpcHealthCheck(context.Background(), client); err != nil {
```
DimitriosNaikopoulos, line 106: Do we need to add a test case for this change? I think we have multiclient tests

deployment/multiclient.go
DimitriosNaikopoulos, line 1: Missing changeset update to create new release

deployment/multiclient.go
bytesizedroll: yep, just a draft for now to see thoughts on the idea, will be handling tests and lint

deployment/multiclient.go
```diff
@@ -63,6 +66,20 @@ type MultiClient struct {
 	chainName   string
 }
 
+// rpcHealthCheck performs a basic health check on the RPC client by calling eth_blockNumber
+func (mc *MultiClient) rpcHealthCheck(ctx context.Context, client *ethclient.Client) error {
+	timeoutCtx, cancel := ensureTimeout(ctx, RPCDefaultHealthCheckTimeout)
```
giogam, line 71: Since you're always passing context.Background(), you can skip ensureTimeout and just use context.WithTimeout directly here.

```suggestion
	timeoutCtx, cancel := context.WithTimeout(ctx, RPCDefaultHealthCheckTimeout)
```

DimitriosNaikopoulos (approved):

PR #94, opened 2025-05-20
[WILL NO MERGE]: [spike]: refactor chains in Environment

graham-chainlink:
An attempt to refactor to remove the coupling of different chains (aptos, solana, evm) with the Environment struct.

Note: This is a spike to see how the new design would look, we would have to implement it in a way to support graceful migration in Chainlink Repo

## Design Choices

- new field under `Environment` call Blockchains which has a private field of map to an interface type replaces all the chain specific fields. No longer we have a new map field for each new chain that we want to support.

```
chains map[uint64]Chain,
solChains map[uint64]SolChain,
aptosChains map[uint64]AptosChain,
```

now represented as

```
chains map[uint64]BlockChain
```
where as `BlockChain` is an interface

Users will be still able to access the specific chain type using the newly provided method, `EVMBySelector`, `SolanaBySelector` etc.

### Benefits

This stops the top level `Environment` struct from containing all the chain fields and grow with every new chain which is not scalable. 

Chain Devs no longer need to update Environment struct to add the new Chain field

Better encapsulation on the chain type, currently in changeset we see usage of `e.SolChain[chainSelector]` or `e.Chain[chainSelector]` everywhere in the changeset, this exposes the underneath data structure as a map which can make things hard to change in future.

---

- Introduce methods `EVMBySelector`, `SolanaBySelector` etc to allow users to get access to the concrete chain type if Blockchain interface is not enough.

In an Ideal world, we want to just work with a single type and standarise all behaviour from all chain families, however this is no feasible as each chain can vary a lot (more reasoning [here](https://chainlink-core.slack.com/archives/C07CF8V02KX/p1744725286124229?thread_ts=1744650825.564109&cid=C07CF8V02KX)).

Which is why the new design is more a hybrid, it works with the single interface type at the `Environment` level but when required, methods are be provided to access those concrete types.

---

- all chains implementations are moved into a new root folder `chain` with each chain type having it owns package.

```
├── chain
│   ├── aptos
│   │   └── aptos.go
│   ├── blockchain.go
│   ├── evm
│   │   └── evm.go
│   └── solana
│       └── solana.go
```

### Benefits

This allows us to assign appropriate product CODEOWNERS to the right chain package. Better ownership control.

Avoid any changes on the core framework code (Environment). To add new chains, only the package `chain` will need to be updated.

--- 

- the existing methods `AllChainSelectorsAllFamiliesExcluding`, `AllChainSelectorsAllFamilies`, `AllChainSelectorsAptos` , `AllChainSelectorsSolana` ,`AllChainSelectors` , `AllChainSelectorsExcluding` are replaced with a single method `ListChainSelectors`  supported by functional options for filtering

```
// WithChainSelectorsExclusion
// WithFamily
func (b BlockChains) ListChainSelectors(options ...ChainSelectorsOption) []uint64 {
```

### Benefit:

The existing design did not scale well, we would have to introduce a new `AllChainSelectors<Chain>` for every new chain we want to support.

Chain dev no longer have to remember to update this file to add the new function

### DIFF
--- a/chain/blockchain.go
+++ b/chain/blockchain.go
@@ -1,0 +1,153 @@
+package chain
+
+import (
+	"errors"
+	"sort"
+
+	chain_selectors "github.com/smartcontractkit/chain-selectors"
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/aptos"
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/evm"
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/solana"
+)
+
+var _ BlockChain = evm.Chain{}
+var _ BlockChain = solana.Chain{}
+var _ BlockChain = aptos.Chain{}
+
+type BlockChains struct {
+	chains map[uint64]BlockChain
+}
+
+type BlockChain interface {
+	String() string
+	Name() string
+	ChainSelector() uint64
+}
+
+func (b BlockChains) EVMChains() (map[uint64]evm.Chain, error) {
+	var evmChains = make(map[uint64]evm.Chain)
+	for selector, chain := range b.chains {
+		c, ok := chain.(evm.Chain)
+		if !ok {
+			continue
+		}
+		evmChains[selector] = c
+	}
+	return evmChains, nil
+}
+
+func (b BlockChains) SolanaChains() (map[uint64]solana.Chain, error) {
+	var solanaChains = make(map[uint64]solana.Chain)
+	for selector, chain := range b.chains {
+		c, ok := chain.(solana.Chain)
+		if !ok {
+			continue
+		}
+		solanaChains[selector] = c
+	}
+	return solanaChains, nil
+}
+
+func (b BlockChains) AptosChains() (map[uint64]aptos.Chain, error) {
+	var aptosChains = make(map[uint64]aptos.Chain)
+	for selector, chain := range b.chains {
+		c, ok := chain.(aptos.Chain)
+		if !ok {
+			continue
+		}
+		aptosChains[selector] = c
+	}
+	return aptosChains, nil
+}
+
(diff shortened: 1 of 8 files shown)


deployment/environment.go
```diff
@@ -187,118 +135,18 @@ func (e Environment) Clone() Environment {
...
-		if slices.Contains(excluding, sel) {
-			continue
-		}
-		ret = append(ret, sel)
-	}
-
-	return ret
-}
-
-func (e Environment) AllDeployerKeys() []common.Address {
```
graham-chainlink, line 290: This does not get used anywhere in the whole company (from compnay wide github search), deleting it.

chain/blockchain.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/evm"
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/solana"
+)
+
+var _ BlockChain = evm.Chain{}
+var _ BlockChain = solana.Chain{}
+var _ BlockChain = aptos.Chain{}
+
+type BlockChains struct {
+	chains map[uint64]BlockChain
```
jkongie, line 19: There are many times where people iterate over the blockchains in a loop, so you may want to consider adding a function like `All()` which returns an iterator

deployment/environment.go
```diff
@@ -44,47 +40,6 @@ type OffchainClient interface {
 	csav1.CSAServiceClient
 }
 
-// Chain represents an EVM chain.
```
jkongie, line 47: In the actual migration, we will want to retain the old chains so that people can upgrade in their own terms. This is mainly because it will involve updating all their changesets as well

graham-chainlink: yes , that is the plan, will maintain the backwards compatibility while we migrate.

PR #98, opened 2025-05-21
deployment/environment.go: add Sui to Environment

cfal:

PR #108, opened 2025-05-22
[CLD-267]: feat: new blockchain field

graham-chainlink:
As part of improving the existing design for chains to be more scalable, we are moving the specific chain code into their own package.

These chains will be abstracted by a higher level field call BlockChains.

This limits the exposure of Map fields externally and allow the design to scale better as we introduce more chains to the framework

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-267


chain/blockchain.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/solana"
+)
+
+var _ BlockChain = evm.Chain{}
+var _ BlockChain = solana.Chain{}
+var _ BlockChain = aptos.Chain{}
+
+type BlockChains struct {
+	chains map[uint64]BlockChain
+}
```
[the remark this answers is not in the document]
graham-chainlink: hmm, it can be, but whats your reasoning? Do you think it will be useful? 

I feel like interface may not be needed as we wont swap around different implementation of BlockChains unlike chains, BlockChains here is really just a wrapper of a map to the Chain interface.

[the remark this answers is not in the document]
graham-chainlink: actually we dont have to even decide here,  since golang is duck typing, we can introduce the interface on the CLD side if we want

chain/common/chain_info_provider.go
```diff
@@ -0,0 +1,56 @@
...
+	chainInfo, err := ChainInfo(c.Selector)
+	if err != nil {
+		return ""
+	}
+	if chainInfo.ChainName == "" {
+		return strconv.FormatUint(c.Selector, 10)
+	}
+
+	return chainInfo.ChainName
+}
```
jkongie, line 37: Let's add comments for these functions to set expectations for others that they should provide details on their code.

chain/common/chain_info_provider.go
```diff
@@ -0,0 +1,56 @@
+package common
+
+import (
+	"fmt"
+	"strconv"
+
+	chain_selectors "github.com/smartcontractkit/chain-selectors"
+)
+
+type ChainInfoProvider struct {
```
jkongie, line 10: Also how do you feel about changing the name to something like `ChainBase`? I think we may use the word `Provider` later elsewhere and it may be confusing.

chain/common/chain_info_provider.go
jkongie, line 1: Could we make this package `internal` to the `chain` package? We want to avoid exposing any packages that are not absolutely required to ensure that people are not using anything they shouldn't and we have to unwind it later.

deployment/environment.go
```diff
@@ -147,6 +107,7 @@ func NewEnvironment(
 	offchain OffchainClient,
 	ctx func() context.Context,
 	secrets OCRSecrets,
+	blockChains chain.BlockChains,
```
jkongie, line 110: I believe this is in a few places. Are you concerned about this being a breaking change?

chain/common/chain_info_provider.go
graham-chainlink: sounds good, i dont mind the name!

deployment/environment.go
graham-chainlink: yeh after some thought , i am going to create a new `NewCLDFEnvironment` to avoid the breaking change, even though i think the [SUI change](https://github.com/smartcontractkit/chainlink-deployments-framework/pull/98) already breaks it :D 

Once we move to my new design, adding a new chain wont break the code anymore

chain/common/chain_info_provider.go
graham-chainlink: cool, i made it public previous because of the method ChainInfo, but i think can make another copy of that function in Chainlink since it is not complicated. Package is now internal

chain/common/chain_info_provider.go
graham-chainlink: done

deployment/environment.go
graham-chainlink: so my plan is 
- to merge [SUI PR](https://github.com/smartcontractkit/chainlink-deployments-framework/pull/98)
- merge this PR
- create new PR to move SUI change into the blockchain field.
- release new version of CLDF

Then this new version wont have breaking change when it gets updated in Chainlink and CLDF

chain/common/chain_info_provider.go
graham-chainlink: After some thinking , i renamed this to ChainMetadata,  I cant call it ChainBase because it cant be a base anymore. 
The reason is i cant do 
```go
type Chain struct {
        ChainBase

	Client OnchainClient
	// Note the Sign function can be abstract supporting a variety of key storage mechanisms (e.g. KMS etc).
	DeployerKey *bind.TransactOpts
}
```
This is a breaking change as now consumer need to do `chain := Chain{ChainBase: ChainBase{Selector:1}....}`, so i opted to do below for each chain
```
// ChainSelector returns the chain selector of the chain
func (c Chain) ChainSelector() uint64 {
	return c.Selector
}

// String returns chain name and selector "<name> (<selector>)"
func (c Chain) String() string {
	return internal.ChainMetadata{Selector: c.Selector}.String()
}

// Name returns the name of the chain
func (c Chain) Name() string {
	return internal.ChainMetadata{Selector: c.Selector}.Name()
}

```
A little duplication but i avoided the breaking change for aptos, sui and evm, and also avoided the clunkiness of setting selector like `chain := Chain{ChainBase: ChainBase{Selector:1}....}`

Usually to get around this , we can create NewEVMChain(selector, ...), but since Chain is a simple struct with all Public fields, we will just introduce an extra layer for not much benefit.

chain/common/chain_info_provider.go
jkongie: yeah I think it's fine to have a `utils` package inside `chains` to contain the `ChainInfo` details as a convenience wrapper

deployment/environment.go
jkongie: Sounds good!

chain/common/chain_info_provider.go
graham-chainlink: cool added a util package under chain

jkongie (approved):

PR #105, opened 2025-05-21
add TON support

huangzhen1997:

PR #110, opened 2025-05-23
feat(chain): add ton support

graham-chainlink:
Migrating [this PR](https://github.com/smartcontractkit/chainlink-deployments-framework/pull/105) raised by Huang to the new chain redesign.


deployment/environment.go
```diff
@@ -270,7 +271,12 @@ func (e Environment) AllChainSelectorsAllFamilies() []uint64 {
 		// ignoring error here to avoid breaking change while we transition to BlockChains.ListChainSelectors
 		suiChains = map[uint64]sui.Chain{}
 	}
-	selectors := make([]uint64, 0, len(e.Chains)+len(e.SolChains)+len(e.AptosChains)+len(suiChains))
+	tonChains, err := e.BlockChains.TonChains()
+	if err != nil {
+		// ignoring error here to avoid breaking change while we transition to BlockChains.ListChainSelectors
+		tonChains = map[uint64]ton.Chain{}
+	}
+	selectors := make([]uint64, 0, len(e.Chains)+len(e.SolChains)+len(e.AptosChains)+len(suiChains)+len(tonChains))
```
jkongie, line 279: I think it would be better to avoid using `Blockchains` until we can ensure that we have all the loaders updated to populate Blockchains

graham-chainlink: yeah i have a [PR populating blockchains](https://github.com/smartcontractkit/chainlink-deployments/pull/3298) over at the CLD side.

This is just more a safe guard, if ton and sui are not populated, then nothing changes.
If they are populated, then atleast this method will return those chains to help with the backwards compatibility while migrating the usage of this method over to the ListChainSelectors

graham-chainlink: Gotcha, we can play it safe,  i removed it.

I do have a [PR in draft](https://github.com/smartcontractkit/chainlink-deployments/pull/3298) for populating the blockchains field

jkongie (approved):

PR #112, opened 2025-05-23
feat: add Family method to the Blockchain interface and implementations

jkongie:
This change introduces a new method, `Family`, to the `Blockchain` interface and updates the chain implementations to return the family name of the respective blockchain.

This allows for better identification of the family rather than relying on the type of the chain itself.

The `ListChainSelectors` method is updated to use this method to simplify it's implementation.


graham-chainlink (approved):
Nice one!

PR #119, opened 2025-05-26
fix(chain): ListChainSelectors filter with multiple families

graham-chainlink:
- Use map instead of slice for faster lookup
- WithFamily now supports more than 1 family, eg `ListChainSelectors(WithFamily("evm"), WithFamily("aptos")`

### DIFF
--- a/chain/blockchain.go
+++ b/chain/blockchain.go
@@ -144,22 +144,32 @@ func (b BlockChains) TonChains() map[uint64]ton.Chain {
 type ChainSelectorsOption func(*chainSelectorsOptions)
 
 type chainSelectorsOptions struct {
-	family    string
-	excluding []uint64
+	// Use map for faster lookups
+	includedFamilies  map[string]struct{}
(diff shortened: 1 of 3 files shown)


chain/blockchain.go
```diff
@@ -144,22 +144,32 @@ func (b BlockChains) TonChains() map[uint64]ton.Chain {
 type ChainSelectorsOption func(*chainSelectorsOptions)
 
 type chainSelectorsOptions struct {
-	family    string
-	excluding []uint64
+	// Use map for faster lookups
+	familiesToInclude  map[string]struct{}
+	chainSelsToExclude map[uint64]struct{}
```
jkongie, line 149: Perhaps this naming would be a bit more concise

```suggestion
	includedFamilies  map[string]struct{}
	excludedChainSels map[uint64]struct{}
```

graham-chainlink: nice one, naming is so hard for me.

jkongie (approved):

PR #123, opened 2025-05-27
feat: adds aptos chain providers

jkongie:


chain/provider.go
```diff
@@ -0,0 +1,8 @@
+package chain
+
+type Provider interface {
+	Initialize() error
```
graham-chainlink, line 4: I wonder instead of having both `Initialize` and `BlockChain` , we could combine them?
Currently the user can call BlockChain without calling Initialize first which can cause user error.

What if we combine them
If Initialize is called , it builds the chain and returns it.
If initialize is called again, it checks if p.Chain is not nil, then returns it if exists.

Thoughts?

chain/internal/common/chain_metadata.go
```diff
@@ -14,6 +14,12 @@ type ChainMetadata struct {
 	Selector uint64
 }
 
+func NewChainMetadata(selector uint64) ChainMetadata {
```
graham-chainlink, line 17: Is this being used somewhere? I didn see any usage

chain/provider.go
graham-chainlink: i do like the fact that initialize is not called in the constructor for cases like CTF where it actually setups the docker which is a bit heavy on construction.

What about moving intialize into BlockChain(), so it can be lazy, chain is initialised when BlockChain is called? and if blockchain is called again, then it just returns the cache.

Thought?

graham-chainlink (approved):

PR #124, opened 2025-05-28
SPIKE: feat(datastore): replace generics metadata with any [DO NOT MERGE]

giogam:
This PR introduces some potential UX improvements to the Datastore API by replacing generics used for custom domain metadata with a more flexible any-based representation provided by Go.

Generics have showed their limitations in the following ways:

- Since the Environment injected into each `Changeset` implementation needs to be consistent across all environments, we can’t easily support custom instantiation code per domain. This adds two extra conversion steps: one at the start of the changeset, and another at the end before passing the datastore to a `ChangesetOutput`.
See [this PR](https://github.dev/smartcontractkit/chainlink/pull/17357/files) for a full example of how a changeset would be implemented using generics.

- Making Environment and `ChangesetOutput` generic is a heavy lift for limited benefit. It would require replacing every `Datastore[DefaultMetadata, DefaultMetadata]` with a fully generic version, pushing type instantiation up a level. That means all `changesets` - even those that don’t need custom metadata - would have to deal with generics, adding unnecessary complexity across the board. Unfortunately, Go doesn’t support generics on methods only, which would have been ideal for a more targeted solution.

The only strong case for keeping generics would be if we could handle environment instantiation at load time and automatically inject a correctly typed Environment into each changeset, which seems very hard to achieve at the moment.

Therefore we propose the following simplification to the current API

**Removal of Generics from the top level API**

While we can still use generics to define interfaces, we propose replacing the generics for contract and environment metadata with `any`. This leads to a leaner API - it eliminates the need for two conversion steps per `changeset`. Users can still work with their own types, with a small trade-off: they'll need to convert from any when reading from the datastore (e.g., after a `Get()`), which can be easily handled using the provided `As` helper that works for any type.

**Universal Clone Logic**

Remove the requirement for users to implement Clone for their custom metadata types. We can just provide a built-in Clone implementation that works generically by leveraging Go’s `JSON` serializability - based on the requirement that all metadata types conform to this standard. Additionally cloning is only relevant for the in-memory Datastore (e.g., an alternative implementation backed by a different data source - such as a database - wouldn’t need it), 

The following is an example of what a changeset looks like before and after applying these changes.

Before
```go
...
// perform conversion from default datastore to domain typed at the beginning of the changeset
envDatastore, err := ds.FromDefault[
	SimpleContract,
	SimpleEnv,
](e.DataStore)


dataStore := ds.NewMemoryDataStore[
	SimpleContract,
	SimpleEnv,
]()

ierr = dataStore.ContractMetadata().Add(
		ds.ContractMetadata[SimpleContract]{
			ChainSelector: 1,
			Address:      "0x1234567890abcdef1234567890abcdef12345678",
			Metadata: SimpleContract{
				DeployedAt:  time.Now(),
				TxHash:      common.HexToHash("0xabcdefabcdefabcdefabcdefabcdefabcdefabcdef"),
				BlockNumber: 123456,
			},
		},

// extra step to convert the typed datastore back to the default one
ds, err := ds.ToDefault(dataStore.Seal())
return deployment.ChangesetOutput{
	DataStore: ds,
}, nil
...
```

After
```go
...
// Conversion steps are not needed anymore

ds := NewMemoryDataStore()

err := ds.ContractMetadata().Add(
	ContractMetadata{
		ChainSelector: 1,
		Address:       "0x1234567890abcdef1234567890abcdef12345678",
		Metadata: SimpleContract{
			DeployedAt:  time.Now(),
			TxHash:      common.HexToHash("0xabcdefabcdefabcdefabcdefabcdefabcdefabcdef"),
			BlockNumber: 123456,
		},
	},
)

record := ds.ContractMetadata().Get(1, "0x1234567890abcdef1234567890abcdef12345678")
// The downside is that when reading values from the datastore, they need to be
// converted back to their original type, since they’re stored as `any`.
typedMeta, err := As[SimpleContract](record.Metadata)
..
```


giogam:
Closing this, as it was only a PoC PR.
The actual changes have been implemented in the following PRs:

- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/141
- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/149
- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/118

giogam:
Closing this, as it was only a PoC PR.
The actual changes have been implemented in the following PRs:

- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/141
- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/149
- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/118

PR #132, opened 2025-05-29
feat: reorder rpcs if they fail all retries

DimitriosNaikopoulos:
## feat: reorder rpcs if they fail all retries

Currently if the default RPC (or any other RPC) starts failing after the client initialisation the multi-client will still keep the same order for RPCs possibly using the same bad RPC on the next call. This change reorders the RPCs and changes the default RPC when a request failed in a way that all RPCs that failed will go down on the RPC's backup list and the successful RPC will become the primary. If the primary RPC was successful or there are no backups, no re-ordering will be performed. 

### Test
Durable Pipeline: https://github.com/smartcontractkit/chainlink-deployments/actions/runs/15415638318

___
Issue:[ CLD-261](https://smartcontract-it.atlassian.net/browse/CLD-261)

### DIFF
--- a/deployment/multiclient.go
+++ b/deployment/multiclient.go
@@ -443,12 +443,12 @@ func ensureTimeout(parent context.Context, timeout time.Duration) (context.Conte
 // If backup RPCs also failed, they will be moved to the end of the backup list.
 // If the primary RPC worked, it will remain the first in the list.
 func (mc *MultiClient) reorderRPCs(rpcIndex int) {
+	mc.mu.Lock()
+	defer mc.mu.Unlock()
+
 	if rpcIndex < 1 || len(mc.Backups) == 0 {
 		return // No need to reorder if the first RPC is still the default or we don't have backups
 	}
-
-	mc.mu.Lock()
-	defer mc.mu.Unlock()
 
 	// Find the index of the backupRPC
 	newDefaultRPCIndex := rpcIndex - 1


deployment/multiclient_test.go
```diff
@@ -295,3 +295,101 @@ func TestEnsureTimeout(t *testing.T) {
...
+		},
+		{
+			name:           "Move last backup to primary",
+			backups:        rpcClients,
+			newDefaultIdx:  3,
+			expectedClient: client3,
+			expectedBackups: []*ethclient.Client{
+				client1,
+				client2,
+				client0,
```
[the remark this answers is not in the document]
DimitriosNaikopoulos: Because the successful RPC was the last RPC on the list, that means that we are just adding the last primary rpc to the end of the backup list. If the primary RPC fails, it will always end up at the end of the backup list.

deployment/multiclient.go
```diff
@@ -431,3 +436,29 @@ func ensureTimeout(parent context.Context, timeout time.Duration) (context.Conte
...
+// If the default RPC failed all attempts, it will be moved to the end of the backup list.
+// If backup RPCs also failed, they will be moved to the end of the backup list.
+// If the primary RPC worked, it will remain the first in the list.
+func (mc *MultiClient) reorderRPCs(rpcIndex int) {
+	if rpcIndex < 1 || len(mc.Backups) == 0 {
+		return // No need to reorder if the first RPC is still the default or we don't have backups
+	}
+
+	mc.mu.Lock()
+	defer mc.mu.Unlock()
```
graham-chainlink, line 450: so we need to add RLock on places where we read mc.Backups and mc.Client, currently this is not sufficient

DimitriosNaikopoulos: Thanks for the feedback!! Updated with read locks on read operations and a write lock for the reorder action that changes the RPC list. Let me know what you think!!

graham-chainlink: yeah my bad haha

deployment/multiclient.go
```diff
@@ -334,6 +336,9 @@ func (mc *MultiClient) WaitMined(ctx context.Context, tx *types.Transaction) (*t
 		}
 	}
 
+	mc.mu.RLock()
+	defer mc.mu.RUnlock()
+
```
graham-chainlink, line 341: I believe this will cause performance issue, the lock will only be unlocked when the select statement below finish which wait for waitMined to finish. This is a call which can take a while.

The mutex will be locked for longer than it needs to be causing performance impact, we should do 
```
 // Only hold the lock while accessing the clients
 mc.mu.RLock()
 clients := append([]*ethclient.Client{mc.Client}, mc.Backups...)
 mc.mu.RUnlock()
```

deployment/multiclient.go
```diff
@@ -355,29 +360,36 @@ func (mc *MultiClient) WaitMined(ctx context.Context, tx *types.Transaction) (*t
 func (mc *MultiClient) retryWithBackups(ctx context.Context, opName string, op func(context.Context, *ethclient.Client) error) error {
 	var err error
 	traceID := uuid.New()
-	for i, client := range append([]*ethclient.Client{mc.Client}, mc.Backups...) {
+
+	mc.mu.RLock()
+	defer mc.mu.RUnlock()
+
```
graham-chainlink, line 366: this will cause a deadlock, 

- mutex is locked by mc.mu.RLock()
- then we call reorderRPCs which tries to lock the mutex if rpcIndex > 0
- both waiting for each other to unlock = deadlock
```
 // Only hold the lock while accessing the clients
 mc.mu.RLock()
 clients := append([]*ethclient.Client{mc.Client}, mc.Backups...)
 mc.mu.RUnlock()
```

deployment/multiclient.go
```diff
@@ -431,3 +437,36 @@ func ensureTimeout(parent context.Context, timeout time.Duration) (context.Conte
...
 	return context.WithTimeout(parent, timeout)
 }
+
+// reorderRPCs reorders the RPCs based on the latest call.
+// If the default RPC failed all attempts, it will be moved to the end of the backup list.
+// If backup RPCs also failed, they will be moved to the end of the backup list.
+// If the primary RPC worked, it will remain the first in the list.
+func (mc *MultiClient) reorderRPCs(rpcIndex int) {
+	mc.mu.Lock()
+	defer mc.mu.Unlock()
```
graham-chainlink, line 447: i think these 2 lines were better after the if statement like what you had before, so it doesnt capture the lock when it is not needed.

graham-chainlink (commented):
looks good

deployment/multiclient.go
DimitriosNaikopoulos: That was the original thinking although on the next line we are reading the struct. Since we are only evaluating the length of the struct though and not is actual elements it maybe ok to access it without lock? Because the number of RPCs will always stay the same 🤔

graham-chainlink: Oh i missed that! actually you are correct! Lets move it back! My bad

PR #137, opened 2025-05-30
fix: Aptos Chain Providers now return a value instead of the pointer

jkongie:
The provider was returning a pointer `*aptos.Chain` which satisfies the `Blockchain` interface return value. However when calling the `AptosChains` method, this does a type assertion against the `aptos.Chain` value type. This resulted in the `AptosChains` method returning 0 chains.

This commit changes the return value of the `RPCChainProvider` and `CTFChainProvider` to return the `aptos.Chain` value instead of a pointer.


chain/aptos/provider/mocks_test.go
jkongie, line 1: I generated this but never used it so am removing it now

jkongie:
I'm not sure this is the direction that we want to go, but this fixes the immediate problem

PR #152, opened 2025-06-04
[CLD-302]: feat(operations): introduce ExecutionOperationN

graham-chainlink:
There is a scenario discovered when an operation with same input has to be run multiple times, eg when deploying the same MCMS contract multiple times so it can be associated with different roles.

This commit introduces `ExecutionOperationN`, it will execute the provided operation N times.

For idempotency behaviour, we went to ensure on reruns, the `ExecutionOperationN` will return the successful prior results in the same order in the slice. That is why we need to remember the order.

Usage
```go
// execute 3 times
linkDeployReports, err := operations.ExecuteOperationN(
			b, DeployLinkOp, deps, operations.EmptyInput{},
			"deploy-multiple-link-contract", 3,
			operations.WithRetry[operations.EmptyInput, EthereumDeps](),
		)
```

Report
```
  {
    "id": "15ac8420-0b59-4293-a188-948ad43643da",
    "definition": {
      "id": "deploy-link-token-op",
      "version": "1.0.0",
      "description": "Deploy LINK Contract Operation"
    },
    "output": "0x9dc98d0a1362d1aa83a10569c45ef4ee3ef97071",
    "input": {},
    "timestamp": "2025-06-04T15:36:02.066618+10:00",
    "error": null,
    "childOperationReports": null,
    "executionSeries": {
      "id": "deploy-multiple-times",
      "order": 1
    }
  }
```

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-302

### DIFF
--- a/operations/execute.go
+++ b/operations/execute.go
@@ -123,29 +123,7 @@ func ExecuteOperation[IN, OUT, DEP any](
 	var err error
 
 	if executeConfig.retryConfig.Enabled {
-		var inputTemp = input
-
-		// Generate the configurable options for the retry
-		retryOpts := executeConfig.retryConfig.Policy.options()
-		// Use the operation context in the retry
-		retryOpts = append(retryOpts, retry.Context(b.GetContext()))
-		// Append the retry logic which will log the retry and attempt to transform the input
-		// if the user provided a custom input hook.
-		retryOpts = append(retryOpts, retry.OnRetry(func(attempt uint, err error) {
-			b.Logger.Infow("Operation failed. Retrying...",
-				"operation", operation.def.ID, "attempt", attempt, "error", err)
-
-			if executeConfig.retryConfig.InputHook != nil {
-				inputTemp = executeConfig.retryConfig.InputHook(attempt, err, inputTemp, deps)
-			}
-		}))
-
-		output, err = retry.DoWithData(
-			func() (OUT, error) {
-				return operation.execute(b, deps, inputTemp)
-			},
-			retryOpts...,
-		)
+		output, err = executeWithRetry(b, operation, deps, input, executeConfig)
 	} else {
 		output, err = operation.execute(b, deps, input)
 	}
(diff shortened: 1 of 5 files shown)


operations/execute.go
```diff
@@ -123,29 +123,7 @@ func ExecuteOperation[IN, OUT, DEP any](
 	var err error
 
 	if executeConfig.retryConfig.Enabled {
-		var inputTemp = input
```
graham-chainlink, line 126: moved this common logic into a separate method for reusability

operations/report.go
```diff
@@ -21,6 +21,18 @@ type Report[IN, OUT any] struct {
...
+	// MultipleExecution is used to track the execution of an operation that was executed multiple times
+	MultipleExecution *MultipleExecution `json:"multipleExecution,omitempty"`
+}
+
+// MultipleExecution is used to track the execution of an operation that was executed multiple times.
+// It contains the unique iD for the MultipleExecution and the order in which it was executed.
+type MultipleExecution struct {
+	// ID is a unique identifier for the multiple execution. Operation with the same MultipleExecutionID
+	// are executed as part of the same multiple execution.
+	ID string `json:"id"`
```
jkongie, line 33: Instead of having to pass in a new ID, could we have it use the cache key instead?

operations/report.go
```diff
@@ -21,6 +21,18 @@ type Report[IN, OUT any] struct {
 	Err       *ReportError `json:"error"`
 	// stores a list of report ID for an operation that was executed as part of a sequence.
 	ChildOperationReports []string `json:"childOperationReports"`
+	// MultipleExecution is used to track the execution of an operation that was executed multiple times
+	MultipleExecution *MultipleExecution `json:"multipleExecution,omitempty"`
```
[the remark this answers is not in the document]
graham-chainlink: I will think about it too, naming is hard for me haha.

What about `ExecutionGroup` ?

operations/execute.go
```diff
@@ -291,3 +376,66 @@ func loadPreviousSuccessfulReport[IN, OUT any](
 	// No previous execution was found
 	return Report[IN, OUT]{}, false
 }
+
+// reportWithOrder allows us to keep track of the order in which reports were executed
+// as part of a multiple execution. This is useful when we want to return the reports in the order
+// they were executed, as the reports are returned in the order they were found in the previous reports.
+type reportWithOrder[IN, OUT any] struct {
```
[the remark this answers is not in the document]
graham-chainlink: actually i just realized i dont need `reportWithOrder` anymore haha, i will remove it.

operations/execute.go
```diff
@@ -166,6 +144,113 @@ func ExecuteOperation[IN, OUT, DEP any](
...
+		"remainingTimesToRun", remainingTimesToRun)
+
+	executeConfig := &ExecuteConfig[IN, DEP]{
+		retryConfig: newDisabledRetryConfig[IN, DEP](),
+	}
+	for _, opt := range opts {
+		opt(executeConfig)
+	}
+
+	results := previousReports
```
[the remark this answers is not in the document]
graham-chainlink: sure i can move things around, make it:
```
resultsLength := uint(len(results))
remainingTimesToRun := n - resultsLength
```
since it makes more sense than below just for readability
```
remainingTimesToRun := n - order
```

operations/execute.go
```diff
@@ -166,6 +144,113 @@ func ExecuteOperation[IN, OUT, DEP any](
...
+// It is important to use a unique executionSeriesID for different sets of multiple executions.
+func ExecuteOperationN[IN, OUT, DEP any](
+	b Bundle, operation *Operation[IN, OUT, DEP], deps DEP, input IN, executionSeriesID string, n uint,
+	opts ...ExecuteOption[IN, DEP],
+) ([]Report[IN, OUT], error) {
+	if !IsSerializable(b.Logger, input) {
+		return []Report[IN, OUT]{}, fmt.Errorf("operation %s input: %w", operation.def.ID, ErrNotSerializable)
+	}
+
+	results, found := loadSuccessfulMultipleExecutionReports[IN, OUT](b, operation.def, input, executionSeriesID)
```
jkongie, line 160: **nit** Perhaps renaming `found` to `ok` may be more conventional

```suggestion
	results, ok := loadSuccessfulMultipleExecutionReports[IN, OUT](b, operation.def, input, executionSeriesID)
```

operations/execute.go
```diff
@@ -166,6 +144,113 @@ func ExecuteOperation[IN, OUT, DEP any](
...
+	if found {
+		// if there are more reports than n, we return only the first n reports
+		if uint(len(results)) >= n {
+			b.Logger.Infow("Operations already executed in multiple execution. Returning previous results", "id", operation.def.ID,
+				"version", operation.def.Version, "description", operation.def.Description, "executionSeriesID", executionSeriesID)
+
+			return results[:n], nil
+		}
+	}
+	resultsLength := uint(len(results))
```
jkongie, line 170: Maybe move this above the `if found` so you can resuse it in the nested conditional.

Consider a rename to `resultsLen` for brevity

operations/execute.go
```diff
@@ -166,6 +144,113 @@ func ExecuteOperation[IN, OUT, DEP any](
 	return report, nil
 }
 
+// ExecuteOperationN executes the given operation multiple n times with the given input and dependencies.
+// Execution will return the previous successful execution results and skip execution if there were
+// previous successful runs found in the Reports.
+// executionSeriesID is used to identify the multiple executions as a single unit.
+// It is important to use a unique executionSeriesID for different sets of multiple executions.
+func ExecuteOperationN[IN, OUT, DEP any](
+	b Bundle, operation *Operation[IN, OUT, DEP], deps DEP, input IN, executionSeriesID string, n uint,
```
jkongie, line 153: **nit** consider shortening this variable name. It has enough context from the function name to drop the `execution` prefix

```suggestion
	b Bundle, operation *Operation[IN, OUT, DEP], deps DEP, input IN, seriesID string, n uint,
```

operations/execute.go
```diff
@@ -291,3 +376,55 @@ func loadPreviousSuccessfulReport[IN, OUT any](
 	// No previous execution was found
 	return Report[IN, OUT]{}, false
 }
+
+func loadSuccessfulMultipleExecutionReports[IN, OUT any](
```
jkongie, line 380: Let's add a docstring for this

operations/execute.go
```diff
@@ -291,3 +376,55 @@ func loadPreviousSuccessfulReport[IN, OUT any](
...
+	b Bundle, def Definition, input IN, multipleExecutionID string,
```
jkongie, line 381: This still references `multipleExecutionID` instead of `seriesID`

operations/execute.go
```diff
@@ -291,3 +376,55 @@ func loadPreviousSuccessfulReport[IN, OUT any](
...
+				continue
+			}
+
+			b.Logger.Debugw(fmt.Sprintf("Previous %s execution found", def.ID), "report_id", report.ID)
+
+			foundReports = append(foundReports, typedReport)
+		}
+	}
+
+	b.Logger.Infof("Found %d reports for MultipleExecutionID %q", len(foundReports), multipleExecutionID)
```
jkongie, line 418:
```suggestion
	b.Logger.Infof("Found %d reports for SeriesID %q", len(foundReports), multipleExecutionID)
```

operations/execute_test.go
```diff
@@ -801,6 +802,481 @@ func Test_ExecuteOperation_Concurrent(t *testing.T) {
 		"Reporter should have %d reports", numGoroutines)
 }
 
+func Test_ExecuteOperationN(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name              string
+		n                 uint
+		multipleExecID    string
```
jkongie, line 811:
```suggestion
		seriesID    string
```

operations/execute.go
graham-chainlink: sure, i thought for private methods, we dont have to as it wont appeared on the go docs, but happy to do it anyway

operations/execute_test.go
graham-chainlink: good catch! looks like i didn push!

jkongie (approved):

PR #154, opened 2025-06-04
feat: provide a SendAndConfirmFunction on the Solana Chain

jkongie:
This PR introduces a new Solana RPC client wrapper (`chain/solana/provider/rpcclient`) to provide enhanced transaction handling, including support for retries, transaction modifiers, and simplified confirmation flows.
    
This RPC Client wrapper code is a modifed version of the the one found in `chainlink-ccip` utils. This allows us to avoid importing the entire `chainlink-ccip` package.
    
https://github.com/smartcontractkit/chainlink-ccip/blob/e52d53e1d7f00babee52e4dc462d9f69fbf1466b/chains/solana/utils/common/transactions.go

### DIFF
--- a/chain/solana/provider/rpc_provider.go
+++ b/chain/solana/provider/rpc_provider.go
@@ -1,15 +1,20 @@
 package provider
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
+	sollib "github.com/gagliardetto/solana-go"
 	solrpc "github.com/gagliardetto/solana-go/rpc"
+	solCommonUtil "github.com/smartcontractkit/chainlink-ccip/chains/solana/utils/common"
 
 	"github.com/smartcontractkit/chainlink-deployments-framework/chain"
 	"github.com/smartcontractkit/chainlink-deployments-framework/chain/solana"
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/solana/provider/rpcclient"
 )
 
 // RPCChainProviderConfig holds the configuration to initialize the RPCChainProvider.
@@ -115,16 +120,33 @@ func (p *RPCChainProvider) Initialize() (chain.BlockChain, error) {
 	}
 
 	// Initialize the Solana client with the provided HTTP RPC URL
-	client := solrpc.New(p.config.HTTPURL)
+	client := rpcclient.New(solrpc.New(p.config.HTTPURL), privKey)
 
 	// Create the Solana chain instance with the provided configuration
 	p.chain = &solana.Chain{
 		Selector:     p.selector,
-		Client:       client,
+		Client:       client.Client,
 		URL:          p.config.HTTPURL,
 		DeployerKey:  &privKey,
 		ProgramsPath: p.config.ProgramsPath,
 		KeypairPath:  keypairPath,
+		SendAndConfirm: func(
+			ctx context.Context, instructions []sollib.Instruction, txMods ...rpcclient.TxModifier,
+		) error {
+			_, err := client.SendAndConfirmTx(ctx, instructions,
+				rpcclient.WithTxModifiers(txMods...),
+				rpcclient.WithRetry(1, 50*time.Millisecond),
+			)
+
+			return err
+		},
+		Confirm: func(instructions []sollib.Instruction, opts ...solCommonUtil.TxModifier) error {
+			_, err := solCommonUtil.SendAndConfirm(
+				context.Background(), client.Client, instructions, privKey, solrpc.CommitmentConfirmed, opts...,
+			)
+
+			return err
+		},
 	}
 
 	return *p.chain, nil
@@ -143,5 +165,5 @@ func (p *RPCChainProvider) ChainSelector() uint64 {
 // BlockChain returns the Aptos chain instance managed by this provider. You must call Initialize
 // before using this method to ensure the chain is properly set up.
 func (p *RPCChainProvider) BlockChain() chain.BlockChain {
-	return p.chain
+	return *p.chain
 }
--- a/chain/solana/provider/rpcclient/client.go
+++ b/chain/solana/provider/rpcclient/client.go
@@ -1,0 +1,305 @@
+package rpcclient
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/avast/retry-go/v4"
+	sollib "github.com/gagliardetto/solana-go"
+	solrpc "github.com/gagliardetto/solana-go/rpc"
+	"github.com/gagliardetto/solana-go/rpc/jsonrpc"
+)
+
+// sendConfig defines the configuration for sending transactions.
+type sendConfig struct {
+	// RetryAttempts determines how many times to retry sending transactions. This applies to all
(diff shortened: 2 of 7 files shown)


chain/solana/provider/rpc_provider.go
```diff
@@ -115,16 +120,33 @@ func (p *RPCChainProvider) Initialize() (chain.BlockChain, error) {
...
 	// Create the Solana chain instance with the provided configuration
 	p.chain = &solana.Chain{
 		Selector:     p.selector,
-		Client:       client,
+		Client:       client.Client,
 		URL:          p.config.HTTPURL,
 		DeployerKey:  &privKey,
 		ProgramsPath: p.config.ProgramsPath,
 		KeypairPath:  keypairPath,
+		SendAndConfirm: func(
```
DimitriosNaikopoulos, line 133: Shouldn't that be part of the `Client`? Something like `solana.Chain.Client.SendAndConfirm(..)`

chain/solana/provider/rpcclient/client.go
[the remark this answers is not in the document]
DimitriosNaikopoulos: For example we already have some generic retry logic that could be used here as well https://github.com/smartcontractkit/chainlink-deployments-framework/blob/93946c53a55a28c8bd26930970283d684273ba6f/deployment/multiclient.go#L357-L390

Maybe `mutliClient` is responsible for defining that generic logic like retry and backup RPCs then each chain implements their own family specific version of it since each client is different

chain/solana/provider/rpcclient/client.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// FixedRetryOpts returns the fixed retry options for confirming transactions.
+func (c *sendConfig) FixedRetryOpts(ctx context.Context) []retry.Option {
+	return []retry.Option{
+		retry.Context(ctx),
+		retry.Attempts(c.FixedRetryAttempts),
+		retry.Delay(c.RetryDelay),
+		retry.DelayType(retry.FixedDelay),
+	}
```
graham-chainlink, line 55: I dont see why we need to have these 2 , the only difference is `retry.Attempts` ?

chain/solana/provider/rpcclient/client.go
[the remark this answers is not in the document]
jkongie: Every chain and their client is so fundamentally different that it is impossible to generalise this. This is the reason why Chains must be implemented differently with different fields.

chain/solana/provider/rpcclient/tx_modifier.go
```diff
@@ -0,0 +1,40 @@
+package rpcclient
+
+import (
+	sollib "github.com/gagliardetto/solana-go"
+	"github.com/smartcontractkit/chainlink-ccip/chains/solana/utils/fees"
+)
+
+// TxModifier is a dynamic function used to flexibly add components to a transaction such as
+// additional signers, and compute budget parameters
+type TxModifier func(tx *sollib.Transaction, signers map[sollib.PublicKey]sollib.PrivateKey) error
```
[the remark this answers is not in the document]
jkongie: These are the only two fields that can be modified before sending a transaction, so I think this is safe.

While a struct may be appropriate, I think it introduces another level of complication.

chain/solana/provider/rpcclient/tx_modifier.go
```diff
@@ -0,0 +1,40 @@
...
+
+func AddSigners(additionalSigners ...sollib.PrivateKey) TxModifier {
```
graham-chainlink, line 12: Do you want to add some go doc comments here for completion sake?

chain/solana/provider/rpc_provider.go
```diff
@@ -115,16 +120,33 @@ func (p *RPCChainProvider) Initialize() (chain.BlockChain, error) {
 	}
 
 	// Initialize the Solana client with the provided HTTP RPC URL
-	client := solrpc.New(p.config.HTTPURL)
+	client := rpcclient.New(solrpc.New(p.config.HTTPURL), privKey)
```
graham-chainlink, line 123: i wonder if we should push this rpcClient creation out of `RPCChainProvider` , so it becomes

```
type RPCClient interface{
   SendAndConfirmTx()
}
func NewRPCChainProvider(..., rpcClient RPCClient) *RPCChainProvider {}
```

then we decouple rpcclient from provider, it allows us to inject a different solana rpc client if that scenario happens in future. But also it allows us to inject a test rpc client which allows us to perform unit testing for `SendAndConfirm` which is not possible now

chain/solana/provider/rpcclient/client.go
```diff
@@ -0,0 +1,310 @@
...
+	// RetryAttempts specifies how many times to retry the send operations. This is used in all
+	// underlying calls to the RPC client.
+	// Setting this value to 0 will cause the operation to retry indefinitely until it succeeds.
+	RetryAttempts uint
+	// RetryDelay specifies the delay between retries.
+	RetryDelay time.Duration
+	// FixedRetryAttempts provides a fixed number of attempts for certain retryable operations.
+	// Currently this is only used for confirming transactions where we want to retry a fixed
+	// number of times independent of the specified retry attempts.
+	FixedRetryAttempts uint
```
graham-chainlink, line 27: From reading the api, it is not obvious the difference between `RetryAttempts` and `FixedRetryAttempts` , even the comments are slightly confusing, both `RetryAttempts` and `FixedRetryAttempts` are setting the number of retries to a "fixed" value. In reality they are being used in different calls.

Should we rename them appropriately?
```suggestion`
	ConfirmTransactionRetryAttempts uint
```

chain/solana/provider/rpcclient/client.go
```diff
@@ -0,0 +1,310 @@
...
+	// TxModifiers is a list of transaction modifiers that can be applied to the transaction before
+	// sending it. This allows for dynamic modifications such as adding additional signers, setting
+	// compute unit limits, or adjusting the compute unit price.
+	TxModifiers []TxModifier
+	// Specifies the commitment level for the transaction. Currently this is only set to Confirmed
+	// and we do not allow modification of this value. We may change this in the future in
```
graham-chainlink, line 33:
```suggestion
	// and we do not allow modification of this value. We may change this in the future if
```

chain/solana/provider/rpcclient/client.go
jkongie: I'm wary of prescribing 'use' to these variable names because it could change in the future. 

However since we don't expose this field out as configurable, it would be ok to named it `ConfirmXXX` because we change it out without impacting consumers.

I'll make a change as see how it looks

chain/solana/provider/rpcclient/client.go
jkongie: Made a bunch of improvements to the doc strings

chain/solana/provider/rpc_provider.go
jkongie: I thought about this but decided against it because we need access to the methods of underlying client (`GetTransaction`, `SendTransactionWithOpts`, etc).

I also wanted to make it simpler for the user so they didn't have to construct an RPCClient themselves to inject. We also run into a problem with a CTF Provider where the user will not be constructing the client themselves because we initialize the container and construct a client for them based on the container details.

graham-chainlink (approved):
Looks good!

PR #155, opened 2025-06-04
feat(datastore) adds more serialization/deserialization tests

giogam:
This adds tests to confirm that ordering is maintained across the JSON `marshal/unmarshal` of address_refs and contract_metadata. This will be expanded to also cover custom metadata fields

### DIFF
--- a/datastore/memory_address_ref_store_test.go
+++ b/datastore/memory_address_ref_store_test.go
@@ -650,7 +650,7 @@ func TestMemoryAddressRefStore_RecordsOrdering(t *testing.T) {
 	require.NoError(t, err, "Failed to unmarshal store")
 
 	// verify the records are in the same order
-	require.Equal(t, len(originalStore.Records), len(unmarshaledStore.Records),
+	require.Len(t, originalStore.Records, len(unmarshaledStore.Records),
 		"number of records should match after unmarshaling")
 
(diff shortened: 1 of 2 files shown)


datastore/memory_address_ref_store_test.go
```diff
@@ -579,3 +580,102 @@ func TestMemoryAddressRefStore_Filter(t *testing.T) {
...
+	var unmarshaledStore MemoryAddressRefStore
+	err = json.Unmarshal(data, &unmarshaledStore)
+	require.NoError(t, err, "Failed to unmarshal store")
+
+	// verify the records are in the same order
+	require.Len(t, originalStore.Records, len(unmarshaledStore.Records),
+		"number of records should match after unmarshaling")
+
+	// compare each record to ensure order is maintained
+	for i, originalRecord := range originalStore.Records {
```
graham-chainlink, line 657: Actually, i prefer the comparison to be done with JSON string, this is more fool proof as that is essentially what we see in CLD

graham-chainlink: Imagine some one implement MarshalJSON and ruin the ordering, this tests still pass but we dont get what we want anymore.

ajaskolski (approved):

PR #174, opened 2025-06-11
protyping on dynamic operation execution

graham-chainlink:


operations/registry.go
```diff
@@ -0,0 +1,27 @@
+package operations
+
+import "errors"
+
+// OperationRegistry is a store for operations that allows retrieval based on their definitions.
+type OperationRegistry[DEP any] struct {
```
RodrigoAD, line 6: Having an fixed `DEP` makes sense in the current state. 

A potential improvement could be to have dynamic dependencies too as part of the framework, where Operations just declare a dependency definition, and there's a Registry with different implementations. The `Execute` should be smart enough to pick the right ones based on the requested dependencies by the Op.

operations/registry_test.go
```diff
@@ -0,0 +1,136 @@
...
+	)
+	// Create registry with untyped operations
+	registry := NewOperationRegistry(stringOp.AsUntyped(), intOp.AsUntyped())
+
+	// Create execution environment
+	b := NewBundle(context.Background, logger.Nop(), NewMemoryReporter())
+
+	inputs := []any{"input1", 42}
+	defs := []Definition{
+		{ID: "string-op", Version: semver.MustParse("1.0.0")},
```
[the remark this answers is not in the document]
graham-chainlink: yes we can do that

operations/registry_test.go
```diff
@@ -0,0 +1,158 @@
...
+	intOp := NewOperation(
+		"int-op",
+		semver.MustParse("1.0.0"),
+		"Echo integer operation",
+		func(e Bundle, deps Deps2, input int) (int, error) {
+			return input, nil
+		},
+	)
+	// Create registry with untyped operations
+	registry := NewOperationRegistry(stringOp.AsUntyped(), intOp.AsUntyped())
```
DimitriosNaikopoulos, line 40: This is more of a question. 

Could `NewOperationRegistry` call `AsUntyped()` to any provided inputs? Then the user doesn't have to remember doing that for every input 🤔

DimitriosNaikopoulos (approved):

ajaskolski (approved):

operations/registry_test.go
graham-chainlink: alright added a helper 

```
         // Create registry with untyped operations by providing optional initial operation
	registry := NewOperationRegistry(stringOp.AsUntyped())

	// An alternative way to register additional operations without calling AsUntyped()
	RegisterOperation(registry, intOp)
```

PR #200, opened 2025-06-25
feature: add a SignHash method to the KMS Client

jkongie:
This adds a new method `SignHash` to the KMS Client interface, allowing signing of arbitrary hashes. This is useful for scenarios where the signing operation does not require a full transaction context, such as for signing zkSync messages.

The existing `GetTransactOpts` method has been updated to use the new `SignHash` method for code resuability.

### DIFF
--- a/chain/evm/provider/kms_signer.go
+++ b/chain/evm/provider/kms_signer.go
@@ -29,6 +29,9 @@ type KMSSigner struct {
 	// kmsKeyID is the ID of the KMS key used for signing. Required to store it on the struct
 	// so we can use it later to sign transactions.
 	kmsKeyID string
+
+	// Cached ECDSA public key derived from the KMS key.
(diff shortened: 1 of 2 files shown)


chain/evm/provider/kms_signer.go
```diff
@@ -74,7 +82,7 @@ func (s *KMSSigner) GetECDSAPublicKey() (*ecdsa.PublicKey, error) {
 	return pubKey, nil
 }
```
giogam, line 83: should `pubKey` be assigned to `ecdsaPublicKey` before returning? I’m not seeing where `pubKey` gets cached otherwise.

jkongie: Yeah good catch. I missed that entirely

giogam (approved):

PR #219, opened 2025-07-10
CLDF Tron Package

eduard-cl:


chain/solana/provider/rpc_provider_test.go
```diff
@@ -33,44 +33,6 @@ func Test_RPCChainProviderConfig_validate(t *testing.T) {
 			giveConfigFunc: func(c *RPCChainProviderConfig) { c.DeployerKeyGen = nil },
 			wantErr:        "deployer key generator is required",
 		},
-		{
```
graham-chainlink, line 36: how come these are deleted in the solana provider?

go.mod
```diff
@@ -9,7 +9,8 @@ require (
 	github.com/aptos-labs/aptos-go-sdk v1.6.3-0.20250331001805-0680b714db6d
 	github.com/avast/retry-go/v4 v4.6.1
 	github.com/aws/aws-sdk-go v1.55.7
-	github.com/ethereum/go-ethereum v1.15.3
+	github.com/ethereum/go-ethereum v1.15.6
+	github.com/fbsobreira/gotron-sdk v0.24.0
```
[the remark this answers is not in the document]
eduard-cl: Would you suggest to use the forked gotron-sdk to include http client? Or is it safe to continue with grpc for CLD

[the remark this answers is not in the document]
eduard-cl: Do you think I should switch to the forked relayer and use HTTP instead? If no third party RPCs support gRPC, we won’t be able to use CLD.

graham-chainlink (commented):
you may want to create a changeset file following instructions [here](https://github.com/smartcontractkit/chainlink-deployments-framework/blob/main/CONTRIBUTING.md#how-to-contribute), so it can be used to create a new release once this PR is merged

scripts/java-tron.conf
```diff
@@ -0,0 +1,366 @@
+# modified from https://github.com/tronprotocol/tron-deployment/blob/master/private_net_config.conf
```
graham-chainlink, line 1: hmm is there no way to tell gotron-sdk that scripts can be found in other places other than root? Having it here doesnt feel intuitive.

jkongie:
The scripts that boot a tron node are an outlier in how we run tests against a node for a chain.

You should utilize the chainlink-testing-framework (or extend it if not available) to bring up a docker container which you can then use in the tests.

Additionally this also allows you to create a CTF Provider for Tron that library consumers could use in their own integration tests

chain/tron/provider/rpcclient/client.go
jkongie, line 1: We generally just put the client wrappers inside the provider package.

Is there any reason why this should be it's own package? If so, do we expect any consumers to actually call use these. If not, we should make the package `internal` to ensure that other people are not using something that we consider to be only internal

chain/tron/keystore/keystore.go
```diff
@@ -0,0 +1,56 @@
...
+	"crypto/ecdsa"
+	"errors"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/fbsobreira/gotron-sdk/pkg/address"
+	"github.com/smartcontractkit/chainlink-common/pkg/loop"
+)
+
+type Keystore struct {
+	Keys map[string]*ecdsa.PrivateKey
```
jkongie, line 14: It would be good to document what the `string` key is referring to. I can see it is the address of the private key by looking at the code further down, but detailing it in the comments will give consumers the information without digging into the code

chain/tron/provider/rpcclient/client.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/tron/keystore"
+
+	"github.com/smartcontractkit/chainlink-tron/relayer/sdk"
+
+	cldf_tron "github.com/smartcontractkit/chainlink-deployments-framework/chain/tron"
+)
+
+// ConfirmRetryOpts returns the retry options for confirming transactions.
+func ConfirmRetryOpts(ctx context.Context, c cldf_tron.ConfirmRetryOptions) []retry.Option {
```
jkongie, line 21: Should this be a private function? I don't see a case where a user may need to call it

chain/tron/provider/rpcclient/client.go
```diff
@@ -0,0 +1,122 @@
...
+	}, retryOpts...)
+
+	if err != nil {
+		return receipt, fmt.Errorf("error confirming transaction: %w", err)
+	}
+
+	return receipt, err
+}
+
+func (c *Client) CheckContractDeployed(address address.Address) error {
```
jkongie, line 115: Please document public functions

chain/tron/provider/rpcclient/client.go
```diff
@@ -0,0 +1,122 @@
...
+	return &Client{
+		Client:   client,
+		Keystore: keystore,
+		Account:  account,
+	}
+}
+
+// SendAndConfirmTx builds, signs, sends, and confirms a transaction.
+// It applies any provided options for retries (delay and attempts) and returns the transaction info
+func (c *Client) SendAndConfirmTx(
```
jkongie, line 49: We should add a unit test for this

chain/tron/provider/rpcclient/client.go
jkongie: We should also make sure we have a unit test for this if possible

chain/tron/tron_chain.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/fbsobreira/gotron-sdk/pkg/http/soliditynode"
+	"github.com/smartcontractkit/chainlink-tron/relayer/sdk"
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/tron/keystore"
+
+	cld_common "github.com/smartcontractkit/chainlink-deployments-framework/chain/internal/common"
+)
+
+// ChainMetadata = generic metadata from the framework
+type ChainMetadata = cld_common.ChainMetadata
```
jkongie, line 18: What is the reason we need to type alias this?

chain/tron/tron_chain.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+}
+
+func DefaultTriggerOptions() TriggerOptions {
+	return TriggerOptions{
+		FeeLimit:            10_000_000,
+		TAmount:             0,
+		ConfirmRetryOptions: DefaultConfirmRetryOptions(),
+	}
+}
```
jkongie, line 85: Ideally document these public functions with reasons why the defaults were chosen and what impact they could have when using them

chain/solana/provider/rpc_provider_test.go
eduard-cl: Added the tests back, were deleted by mistake

chain/tron/provider/rpcclient/client.go
eduard-cl: I followed the same structure we use in Solana, where the `provider` contains an `rpcclient` responsible for sending and confirming transactions. I considered splitting it into a separate package, but decided to keep the consistency with the existing structure. That said, if you prefer, I can separate it.

scripts/java-tron.conf
eduard-cl: With the last commits we use CTF instead of scripts to setup a local network

chain/tron/keystore/keystore.go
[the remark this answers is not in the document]
eduard-cl: Added comments to all the non-testing files.

chain/tron/keystore/keystore.go
eduard-cl: Absolutely agree — I’ve added that clarification, along with detailed comments throughout the entire file to improve readability.

chain/tron/provider/rpcclient/client.go
eduard-cl: Yes, modified to private.

chain/tron/provider/rpcclient/client.go
eduard-cl: Added detailed comments throughout the entire file and created some tests deploying a local network.

chain/tron/provider/rpcclient/client.go
eduard-cl: Same as above, unit tests created setting up a local network

chain/tron/tron_chain.go
eduard-cl: I copied other packages like Ton or Sui, they also include this type alias, should I remove it?

chain/tron/tron_chain.go
eduard-cl: Comments added explaining the public functions and chosen values.

chain/tron/tron_chain.go
graham-chainlink: ya,  we will need it as the actual ChainMetadata lives in a common package under internal, it feels more intuitive to be able to do cldf_tron.ChainMetadata from the user perspective.

chain/tron/tron_chain.go
```diff
@@ -0,0 +1,91 @@
...
+type TriggerOptions struct {
+	FeeLimit            int32               // Max TRX to be used for this transaction call.
+	TAmount             int64               // Amount of TRX to transfer along with the contract call (like msg.value).
+	ConfirmRetryOptions ConfirmRetryOptions // Retry options for confirming the transaction.
+}
+
+// Chain represents a Tron chain
+type Chain struct {
+	ChainMetadata                    // Chain selector and metadata
+	Client        sdk.CombinedClient // Combined client for Tron operations
```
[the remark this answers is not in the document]
eduard-cl: Sorry, I don't understand the question, aren't we creating the rpc client for them in the rpc provider?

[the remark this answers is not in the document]
graham-chainlink: ah let me clarify, not sure if it is better or not, you would have better idea of how it is being used.

so currently we have the chain 
```
type Chain struct {
	ChainMetadata                    // Chain selector and metadata
	Client        sdk.CombinedClient // Combined client for Tron operations
```
I was suggesting why not return the rpcclient instead of the combined client
```
type Chain struct {
	ChainMetadata                    // Chain selector and metadata
	Client        rpcclient.Client
```

chain/tron/provider/ctf_provider_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	"github.com/fbsobreira/gotron-sdk/pkg/address"
+	chain_selectors "github.com/smartcontractkit/chain-selectors"
+	"github.com/smartcontractkit/chainlink-evm/gethwrappers/shared/generated/link_token"
+	"github.com/smartcontractkit/chainlink-testing-framework/lib/logging"
```
graham-chainlink, line 11: I would use `t.Logf` from the native testing framework instead of the logger from CTF. Same for the other tests too

chain/tron/provider/rpc_provider_test.go
```diff
@@ -0,0 +1,377 @@
...
+	"testing"
+	"time"
+
+	"github.com/avast/retry-go/v4"
+	"github.com/fbsobreira/gotron-sdk/pkg/address"
+	"github.com/rs/zerolog"
+	chain_selectors "github.com/smartcontractkit/chain-selectors"
+	"github.com/smartcontractkit/chainlink-evm/gethwrappers/shared/generated/link_token"
+	"github.com/smartcontractkit/chainlink-testing-framework/framework/components/blockchain"
+	"github.com/smartcontractkit/chainlink-testing-framework/lib/logging"
```
graham-chainlink, line 15: here to refer to my other comment about using t.Log instead

chain/tron/tron_chain.go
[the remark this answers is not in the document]
graham-chainlink: this is so for example in the ctf_provider_tests.go , we dont have to do wrap it with the rpclient [here](https://github.com/smartcontractkit/chainlink-deployments-framework/blob/3d908bc0c2d591824193417df4896dc32ed35860/chain/tron/provider/ctf_provider_test.go#L191-L191)

```
	tronChain, ok := chainInstance.(tron.Chain)
	require.True(t, ok, "Expected TRON chain instance")

	logger.Info().Str("chainURL", tronChain.URL).Uint64("selector", tronChain.Selector).Msg("TRON CTF chain initialized")

	// Create RPC client using the chain's components
	rpcClient := rpcclient.New(tronChain.Client, tronChain.Keystore, tronChain.Address)
```
we will just have rpcClient directly on the chain instance itself, does that make sense?

chain/tron/provider/ctf_provider_test.go
```diff
@@ -0,0 +1,270 @@
...
+	chainID, err := chain_selectors.GetChainIDFromSelector(chain_selectors.TRON_MAINNET.Selector)
+	require.NoError(t, err)
+	fullNodeURL, solidityNodeURL := provider.startContainer(chainID)
+	require.NotEmpty(t, fullNodeURL)
+	require.NotEmpty(t, solidityNodeURL)
+	require.Contains(t, fullNodeURL, "/wallet")
+	require.Contains(t, solidityNodeURL, "/walletsolidity")
+}
+
+func TestCTFProvider_SendAndConfirmTx_And_CheckContractDeployed(t *testing.T) {
```
graham-chainlink, line 168: the test is include "SendAndConfirm" , but seems like the `SendAndConfirm` from chain created by CTF Provider is not tested at all here

chain/tron/provider/ctf_provider_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// Log the address used to deploy contracts (chain address)
+	t.Logf("Using chain address: chain address=%s", tronChain.Address.String())
+
+	// Generate a random minter address
+	minterAddress, err := address.Base58ToAddress("TQtWBxe8wNAcio3evcfwMAqsdFzykpi6e7")
+	require.NoError(t, err, "Failed to generate minter address")
+
+	// Check the minter role status before granting it
+	beforeMinterResp, err := tronChain.Client.TriggerConstantContract(
```
graham-chainlink, line 211: how come in above we use `tronChain.DeployContractAndConfirm` , but here we use `tronChain.Client.TriggerConstantContract` ? Should we be using `tronChain.DeployContractAndConfirm` ?

eduard-cl: If you go to the provider `Initialize` method you'll see the provider have the following helper methods:
- SendAndConfirm: to send a transaction and confirm it.
- DeployContractAndConfirm: to deploy a contract and confirm it.
- TriggerContractAndConfirm: to send a transaction to a smart contract and confirm it.

The method you're pointing (`tronChain.Client.TriggerConstantContract`) is from the client, to call a view smart contract method, since there's no need to send a transaction, there's no incentive to wrap it in a helper method.

graham-chainlink: ops i just realized my original message is incorrect, i meant `tronChain.TriggerContractAndConfirm` and `tronChain.Client.TriggerConstantContract`  are both used here and why not just use `tronChain.TriggerContractAndConfirm` ?

graham-chainlink (approved):
Lets go!

chain/tron/provider/ctf_provider_test.go
eduard-cl: The `TriggerContractAndConfirm` is a wrap method to simplify two calls in just one:
- `TriggerSmartContract`: To create a Smart Contract transaction.
- `SendAndConfirm`: To send the transaction and confirm using the rpc client.
The `tronChain.Client.TriggerConstantContract` is used to call a view method from a smart contract, we don't need to send a transaction or wait for any confirmation. It would not make sense to wrap it in a helper method.

graham-chainlink (approved):

PR #229, opened 2025-07-15
feat: adds catalog datastore

giogam:


datastore/catalog/contract_metadata_store.go
```diff
@@ -220,38 +224,72 @@ func (s *CatalogContractMetadataStore) Fetch() ([]datastore.ContractMetadata, er
...
 
-func (s *CatalogContractMetadataStore) Add(record datastore.ContractMetadata) error {
-	return s.editRecord(record, pb.EditSemantics_SEMANTICS_INSERT)
+func (s *CatalogContractMetadataStore) Add(ctx context.Context, record datastore.ContractMetadata) error {
+	return s.editRecord(ctx, record, pb.EditSemantics_SEMANTICS_INSERT)
 }
 
-func (s *CatalogContractMetadataStore) Upsert(record datastore.ContractMetadata) error {
-	return s.editRecord(record, pb.EditSemantics_SEMANTICS_UPSERT)
+func (s *CatalogContractMetadataStore) Upsert(ctx context.Context, key datastore.ContractMetadataKey, metadata any, updaters ...datastore.MetadataUpdaterF) error {
```
graham-chainlink, line 245: I wonder instead of having `updaters ...datastore.MetadataUpdaterF` as the last parameter, we could have a generic option object that also provides WithMetadataUpdaterF so in future if we want to provide other configs/options, we can still do that, currently we will lock ourselves to only have `MetadataUpdaterF` option if that make sense? Thoughts?

giogam: Yes, that was actually my initial idea, but I ended up going with the function directly. Let me try to pull something together.

giogam: Pushed some changes let me know if that's what you had in mind!

datastore/catalog/chain_metadata_store.go
```diff
@@ -217,38 +221,73 @@ func (s *CatalogChainMetadataStore) Fetch() ([]datastore.ChainMetadata, error) {
...
+	}
+
+	// Apply user-provided options
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	// Get current record for update operations
+	var currentMetadata any
+	if currentRecord, err := s.Get(ctx, key); err == nil {
```
graham-chainlink, line 267: we didn fully handle the error here, missing a case where there is an error and the error is not `ErrChainMetadataNotFound`

datastore/catalog/contract_metadata_store.go
```diff
@@ -220,38 +225,72 @@ func (s *CatalogContractMetadataStore) Fetch() ([]datastore.ContractMetadata, er
...
+	}
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	// Get current record for update operations
+	var currentMetadata any
+	if currentRecord, err := s.Get(ctx, key); err == nil {
+		currentMetadata = currentRecord.Metadata
+	} else if !errors.Is(err, datastore.ErrContractMetadataNotFound) {
```
graham-chainlink, line 271: same here too missing the catch all error handling

datastore/catalog/env_metadata_store.go
```diff
@@ -153,30 +158,52 @@ func (s *CatalogEnvMetadataStore) Get() (datastore.EnvMetadata, error) {
...
+
+	// Apply user-provided options
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	currentRecord, err := s.Get(ctx)
 	if err != nil {
-		return fmt.Errorf("failed to create gRPC stream: %w", err)
+		if !errors.Is(err, datastore.ErrEnvMetadataNotSet) {
```
graham-chainlink, line 174: here as well missing generic error handling

datastore/catalog/datastore.go
```diff
@@ -2,13 +2,12 @@ package catalog
...
 	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
-	pb "github.com/smartcontractkit/chainlink-deployments-framework/datastore/catalog/internal/protos"
 )
 
 type CatalogDataStoreConfig struct {
-	Domain      string                        `json:"domain"`
-	Environment string                        `json:"environment"`
-	Client      pb.DeploymentsDatastoreClient `json:"-"`
+	Domain      string        `json:"domain"`
+	Environment string        `json:"environment"`
```
graham-chainlink, line 9: Do you know the use of Domain and Env? I dont see any usages?

giogam: They are propagated through the different sections of the store to be used in requests, but not actually used at the top level. I've removed the storing for now.

PR #241, opened 2025-08-05
[CLD-486]: feat: new CTF Provider for JD

graham-chainlink:
Introduce new CTF provider for JD which provides a ready for testing JD docker container.

Because JD docker image is not available publicly, we need to use ECR auth to gain access to the docker image, the ctf_integration_test.go which spins up the JD containers and execute requests on it is currently skipped because of it.  Or just build it locally. 

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-486

### DIFF
--- a/offchain/jd/doc.go
+++ b/offchain/jd/doc.go
@@ -10,14 +10,15 @@ mechanisms and provides a unified interface for job management operations.
 
 # Architecture
 
-The package consists of two main components:
+The package consists of three main components:
 
-1. **JD Client** (`client.go`) - Core gRPC client implementation
-2. **Provider Interface** (`provider/`) - A wrapper around the JD client that provides a standarized interface
+1. JD Client (client.go) - Core gRPC client implementation
+2. Client Provider (provider/client_provider.go) - Connects to existing JD services
+3. CTF Provider (provider/ctf_provider.go) - Creates and manages JD Docker containers for testing
 
 # Basic Usage
 
-## Simple Connection
+# Simple Connection
 
 For basic connectivity without authentication:
 
@@ -38,9 +39,10 @@ For basic connectivity without authentication:
 	}
 
 	// Use client for operations
-	pubKey, err := client.GetCSAPublicKey(ctx)
+	import csav1 "github.com/smartcontractkit/chainlink-protos/job-distributor/v1/csa"
+	keypairs, err := client.ListKeypairs(ctx, &csav1.ListKeypairsRequest{})
 
-## Provider Interface
+# Provider Interface
 
 	import (
 		"github.com/smartcontractkit/chainlink-deployments-framework/offchain/jd/provider"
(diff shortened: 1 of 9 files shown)


offchain/jd/doc.go
```diff
@@ -101,21 +172,38 @@ For services requiring OAuth2 Bearer tokens:
...
 # Client Operations
 
-The JD client supports various operations:
+The JD client supports various operations through gRPC service interfaces:
 
 ## CSA Key Management
 
-	// Get CSA public key
-	pubKey, err := client.GetCSAPublicKey(ctx)
+	import csav1 "github.com/smartcontractkit/chainlink-protos/job-distributor/v1/csa"
```
graham-chainlink, line 179: i updated the docs a bit here too

offchain/jd/provider/ctf_provider.go
```diff
@@ -0,0 +1,173 @@
...
+	ctf_jd "github.com/smartcontractkit/chainlink-testing-framework/framework/components/jd"
+	"github.com/smartcontractkit/chainlink-testing-framework/framework/components/postgres"
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/deployment"
+	"github.com/smartcontractkit/chainlink-deployments-framework/offchain"
+	"github.com/smartcontractkit/chainlink-deployments-framework/offchain/jd"
+)
+
+// CTFOffchainProviderConfig holds the configuration to initialize the CTFOffchainProvider.
+type CTFOffchainProviderConfig struct {
```
jkongie, line 21:
```suggestion
type CTFProviderConfig struct {
```

This is inline with the onchain provider naming

offchain/jd/provider/ctf_provider.go
```diff
@@ -0,0 +1,173 @@
...
+	return client, nil
+}
+
+// healthCheck verifies that the JD service is ready by calling ListKeypairs with retry logic.
+func (p *CTFOffchainProvider) healthCheck(ctx context.Context, client deployment.OffchainClient) error {
+	p.t.Helper()
+
+	err := retry.Do(func() error {
+		// Try to list CSA keypairs as a health check
+		_, err := client.ListKeypairs(ctx, &csav1.ListKeypairsRequest{})
```
jkongie, line 143: There is actually a healthcheck endpoint that you could use for this instead

https://docs.jd.cldev.sh/#/api/app/healthcheck

offchain/jd/doc.go
```diff
@@ -68,6 +70,75 @@ For basic connectivity without authentication:
...
+
+	export CTF_JD_IMAGE=localhost:5001/job-distributor:latest
+
+	config := provider.CTFOffchainProviderConfig{
+		// Image field can be omitted when CTF_JD_IMAGE is set
+	}
+
+### Health Check
+
+The CTF provider includes built-in health checking that retries `ListKeypairs` calls:
```
jkongie, line 138: This will need to be changed if you change to use the Healthcheck call

offchain/jd/provider/ctf_provider.go
graham-chainlink: yeah we could but using the `ListKeypairs` also checks the JD client is setup correct as well as the DB is configured correctly which the health endpoint doesnt check

offchain/jd/provider/client_provider.go
```diff
@@ -39,7 +39,7 @@ var _ offchain.Provider = (*ClientOffchainProvider)(nil)
 // ClientOffchainProvider is a JD provider that connects to a Job Distributor service via gRPC.
 type ClientOffchainProvider struct {
 	config ClientOffchainProviderConfig
-	client deployment.OffchainClient
+	client offchain.Client
```
graham-chainlink, line 42: using the correct package instead of the old `deployment.OffchainClient`

offchain/jd/provider/ctf_provider.go
jkongie: Should we use [`GetKeypair`](https://docs.jd.cldev.sh/#/api/csa/retrieve-csa-keypair) instead then since a JD can only have 1 keypair?

graham-chainlink: Sure updated! i dont think the call matters since we just want to verify db access.

jkongie (approved):

PR #246, opened 2025-08-06
feat: add network config to cld engine

jkongie:
This adds the new structures to represent network configuration. Configuration is loaded from a YAML file, and can be filtered and transformed.

These networks are used to configure the CLD environment.


jkongie:
Updated and ready for a re-review @graham-chainlink

engine/cld/network/config.go
```diff
@@ -0,0 +1,188 @@
+package network
+
+import (
+	"fmt"
+	"maps"
+	"os"
+	"slices"
+
+	"gopkg.in/yaml.v3"
```
graham-chainlink, line 9: Maybe out of scope, but we should use https://github.com/goccy/go-yaml over `gopkg.in/yaml.v3` since the later is an archived repo

graham-chainlink (approved):

PR #251, opened 2025-08-07
migrate(file-json-utils): from cldf to cld

ajaskolski:
migrate json file utils from cld -> cldf


graham-chainlink (approved):
we could probably flatten fileutils and jsonutils into just under lib directly, but i dont mind this too. Shorten import string and i don think those fileutils and jsonutils package will grow much to be in its own package

eg 
lib/fileutils.go
lib/jsonutils.go

PR #261, opened 2025-08-12
refactor(engine): move domain go with minimal dependenciesn from cldf

ajaskolski:


engine/cld/domain/adressbook.go
jkongie, line 1: typo in filename

PR #267, opened 2025-08-13
ci: optimize runner & test config

graham-chainlink:
- bump runner for test to `ubuntu20.04-4cores-16GB` which is twice the number of vCPUs and double the memory at 17GB
- set timeout for test to 10 min so it does not hang indefinitely.
- set go test config -p 2 -parallelism 2 which means at any time run 2 tests over max of 2 different packages. This limits the number of docker containers running at any time.
 
I did a bunch of testing and find this give the best stablilty and performance without going for more expensive runner such as `ubuntu-latest-8-cores`.

This test still runs under 5 minutes and we no longer have the test hanging for ages.


.github/workflows/pull-request-main.yml
```diff
@@ -42,14 +43,15 @@ jobs:
           # disable the checkptr runtime check due a false positive in github.com/xssnick/tonutils-go
           # causing tests in ci to fail "fatal error: checkptr: pointer arithmetic result points to invalid allocation"
           # https://github.com/xssnick/tonutils-go/issues/310
-          go-test-cmd: go test -race -gcflags=all=-d=checkptr=0 -coverprofile=coverage.txt $(go list ./...)
+          # -p 2 -parallel 3 = 2 packages, 3 tests max = 6 containers max
+          go-test-cmd: go test -race -gcflags=all=-d=checkptr=0 -p 2 -parallel 3 -coverprofile=coverage.txt $(go list ./...)
```
giogam, line 47: should we also update https://github.com/smartcontractkit/chainlink-deployments-framework/blob/c01a140379a855cc2300d093b773d60c48da52d0/.github/workflows/schedule-main.yml#L44-L45?

PR #270, opened 2025-08-14
feat: Replace WithRPCURLTransformer with targeted transformers

jkongie:
The previous `WithRPCURLTransformer` load option was a single function that transformed both HTTP and WS URLs. This is too general and makes the assumption that the caller wants to transform all the URLs with the same function.

This PR introduces 2 new load options: `WithHTTPURLTransformer` and `WithWSURLTransformer` which are more specific and allow the caller to transform only or both the HTTP or WS URLs, with different functions.

### DIFF
--- a/engine/cld/config/network/config.go
+++ b/engine/cld/config/network/config.go
@@ -134,13 +134,12 @@ func ChainFamilyFilter(chainFamily string) NetworkFilter {
 	}
 }
 
-// transformRPCURLs transforms the RPC URLs of the networks in the config.
-func (c *Config) transformRPCURLs(transform URLTransformer) {
+// transformHTTPURLs transforms the HTTP URLs of the networks in the config.
+func (c *Config) transformHTTPURLs(transform URLTransformer) {
 	for k, n := range c.networks {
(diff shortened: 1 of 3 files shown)


engine/cld/config/network/config.go
```diff
@@ -187,29 +202,40 @@ func Load(filePaths []string, opts ...LoadOption) (*Config, error) {
...
+		cfg.transformWSURLs(loadCfg.WSURLTransformer)
 	}
 
 	return cfg, nil
 }
 
-// LoadOption defines a function that modifies the load configuration.
-type LoadOption func(*loadConfig)
+// loadOption defines a function which modifies the load configuration.
+type loadOption func(*loadConfig)
```
graham-chainlink, line 218: This should be public right? Else it may be hard for consumer to work with option
```suggestion
// LoadOption defines a function which modifies the load configuration.
type LoadOption func(*loadConfig)
```

jkongie: Thats a good point. I'll change it back.

Thanks

PR #273, opened 2025-08-14
chore: add more tests for the Job Distributor client

jkongie:
This PR adds more tests for the Job Distributor client, introducing mocks for the Job Distributor GRPC clients.


offchain/jd/client_test.go
```diff
@@ -79,3 +83,252 @@ func TestNewJDClient_ConfigurationScenarios(t *testing.T) {
...
+				assert.Empty(t, got)
+			} else {
+				require.NoError(t, err)
+				assert.Equal(t, tt.want, got)
+			}
+		})
+	}
+}
+
+func TestJDClient_ProposeJob(t *testing.T) {
```
DimitriosNaikopoulos, line 198: Do we also want a test for the `NodeServiceClient` ?

jkongie: I didn't see any method there that called it

PR #278, opened 2025-08-15
Create an in-memory catalog implementation for testing.

cgruber:
This uses an in-memory instance of postgresql to fully mirror the functionality of the catalog service (including transactionality without locking up the whole store, etc.)

It uses very basic sql.* pkg code, without an ORM, for simplicity. Each table store (address_reference, etc.) has a set of queries it uses in constants, with some minimal infrastructure factored out. This could be cleaner with more infrastructure or an ORM, keeping it thin here, was important for the first draft. 

Cleaning up resources (`store.Close()`) is very important to do, so postgresql cleans up shared memory resources it makes use of.  In CI pipelines, this may not be so bad, as it takes some time for the shmem references to build up, but a very long set of tests that don't clean up could end up running out of shmem references. While you can always up the container's shmem stats, it's better to just add `defer store.Close()` where you create the store (examples in the tests for this implementation).


giogam (commented):
all comments addressed!

bytesizedroll (approved):

PR #288, opened 2025-08-19
Add Sui env config

RodrigoAD:
- Adds Sui in config

### DIFF
--- a/engine/cld/config/env/config.go
+++ b/engine/cld/config/env/config.go
@@ -49,6 +49,14 @@ type AptosConfig struct {
 	DeployerKey string `mapstructure:"deployer_key"` // Secret: The private key of the deployer account.
 }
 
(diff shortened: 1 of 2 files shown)


engine/cld/config/env/config.go
```diff
@@ -49,6 +49,14 @@ type AptosConfig struct {
 	DeployerKey string `mapstructure:"deployer_key"` // Secret: The private key of the deployer account.
 }
 
+// SuiConfig is the configuration for the Aptos Chains.
```
jkongie, line 52:
```suggestion
// SuiConfig is the configuration for the Sui Chains.
```

jkongie (commented):
Does SUI support ECDSA keys? 

If so, you will also need to support KMS keys because Security requires this for production deploys

PR #292, opened 2025-08-19
Fixes: Sui chain & ctf provider fixes

stackman27:


chain/sui/signer.go
```diff
@@ -46,24 +46,69 @@ func NewSignerFromHexPrivateKey(hexPrivateKey string) (SuiSigner, error) {
...
+	}
+	pub := make([]byte, ed25519.PublicKeySize)
+	copy(pub, priv[32:]) // last 32 bytes are the pubkey
+
+	return pub, nil
+}
+
+// PrivateKey returns the underlying ed25519.PrivateKey (64 bytes = seed||pubkey)
+// from a SuiSigner created in this package.
+func PrivateKey(s SuiSigner) (ed25519.PrivateKey, error) {
```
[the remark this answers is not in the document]
stackman27: hmm any opposition in just retrieving the signer, i need it in the format that blockvision returns

chain/sui/provider/ctf_provider.go
```diff
@@ -204,13 +209,15 @@ func (p *CTFChainProvider) startContainer(
...
+
 			return containerResult{}, rerr
 		}
 
 		testcontainers.CleanupContainer(p.t, output.Container)
 
 		return containerResult{
-			url:           output.Nodes[0].ExternalHTTPUrl + "/v1",
+			url:           output.Nodes[0].ExternalHTTPUrl,
+			fauceturl:     input.FaucetPort,
```
graham-chainlink, line 220:
```suggestion
			faucetPort:     input.FaucetPort,
```

chain/sui/provider/ctf_provider.go
```diff
@@ -240,14 +247,24 @@ func (p *CTFChainProvider) startContainer(
 	}
 	require.True(p.t, ready, "Sui network not ready")
 
-	dc, err := framework.NewDockerClient()
+	err = fundAccount(fauceturl, address)
```
graham-chainlink, line 250: how come we need to do this again? CTF alredy does this underneath right?

stackman27: this is for the deployer key, the rest if for the node transmitter addresses

chain/sui/signer.go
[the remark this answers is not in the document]
graham-chainlink: hmm but seems like the signer has the PriKey as exported , which may be an issue if we return the signer, any alternative like performing the signing here instead of passing the private key out?

chain/sui/signer.go
```diff
@@ -46,24 +46,68 @@ func NewSignerFromHexPrivateKey(hexPrivateKey string) (SuiSigner, error) {
...
+		return nil, fmt.Errorf("unexpected ed25519 key length: %d", len(priv))
+	}
+	pub := make([]byte, ed25519.PublicKeySize)
+	copy(pub, priv[32:]) // last 32 bytes are the pubkey
+
+	return pub, nil
+}
+
+// GetSigner returns the signer in blockVision signer.Signer interface format
+func GetSigner(s SuiSigner) (*signer.Signer, error) {
```
graham-chainlink, line 106: Not blocking this PR, but just raising that returning the signer here exposes the private key which is not something we should do.

graham-chainlink (dismissed):

PR #294, opened 2025-08-20
refactor(engine): move chains & minor env dependencies from cld

ajaskolski:
Extraction of chain loading logic.
https://smartcontract-it.atlassian.net/browse/CLD-476


engine/cld/domain/fixture_test.go
```diff
@@ -27,7 +27,7 @@ type testDomainFS struct {
 
 // TODO: unexport after migrations/archive.go is extracted from cldf
```
DimitriosNaikopoulos, line 28: Should we remove this reminder now?

engine/cld/network/networks.go
```diff
@@ -0,0 +1,134 @@
...
+	// RpcsCldevSh is the host for the CL RPC Proxies.
+	RpcsCldevSh = "rpcs.cldev.sh"
+	// GapRPCProxy is the host for the GAP V2 proxy which is required to be used when accessing
+	// CL RPC Proxies from Github.
+	GapRPCProxy = "gap-rpc-proxy.public.main.prod.cldev.sh"
+	// GapRPCProxyWSPort is the port for the GAP V2 proxy WebSocket endpoint.
+	GapRPCProxyWSPort = "9443"
+	// GapRPCProxyHTTPPort is the port for the GAP V2 proxy HTTP endpoint.
+	GapRPCProxyHTTPPort = "4443"
+)
```
DimitriosNaikopoulos, line 27: Are we ok with moving these endpoints to the public repo?

engine/cld/network/networks.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// If we are in CI, we want to use the GAP DNSs for the RPC URLs.
+	// TODO: This is a temporary solution as this code will soon be moved to the framework. We
+	// should take into account that reading the env var is not a great solution, as it can lead
+	// to unexpected behavior based on an declared env var buried in code.
+	//
+	// Instead, we should consider moving the fetching of the env var into the commands that need
+	// it, and determine the correct behavior based on the command.
+	var loadOpts []cldf_config_network.LoadOption
+	if os.Getenv("GITHUB_ACTIONS") == "true" {
```
DimitriosNaikopoulos, line 113: Same question about moving this to the public repo

engine/cld/domain/fixture_test.go
ajaskolski: Yes good catch

ajaskolski: deleted

engine/cld/chains/chains.go
```diff
@@ -0,0 +1,553 @@
...
+}
+
+// Load loads an Aptos Chain for a selector.
+func (l *chainLoaderAptos) Load(ctx context.Context, selector uint64) (cldf_chain.BlockChain, error) {
+	network, err := l.getNetwork(selector)
+	if err != nil {
+		return nil, err
+	}
+
+	rpcURL := network.RPCs[0].HTTPURL
```
bytesizedroll, line 248: loading the first in each chain is fine right? because the getNetwork handles the fallback and rearrangement?

bytesizedroll (approved):

engine/cld/chains/chains.go
DimitriosNaikopoulos: getNetwork only returns the network config. Because this is an Aptos Client I don't think we have a fallback logic 🤔 Only EVMs are using the multi client which has the fallback and RPC reshuffling logic

ajaskolski: Yes its the same for Solana for example.

PR #297, opened 2025-08-20
feat: add yaml tags to env config allowing for direct yaml file loading

jkongie:
This change adds yaml tags to the env config allowing for direct yaml file loading.

### DIFF
--- a/.changeset/fine-crews-hope.md
+++ b/.changeset/fine-crews-hope.md
@@ -1,0 +1,5 @@
+---
+"chainlink-deployments-framework": minor
+---
+
+Allow users to marshal and unmarshal env config directly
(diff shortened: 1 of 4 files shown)


graham-chainlink (commented):
do you want to have a test for loading a yaml file into those struct?

jkongie:
> do you want to have a test for loading a yaml file into those struct?

I wasn't planning to since it's so simple, but I can if you think there is value

graham-chainlink:
> > do you want to have a test for loading a yaml file into those struct?
> 
> I wasn't planning to since it's so simple, but I can if you think there is value

@jkongie  we can keep it super simple one test just tries to unmarshal a yaml into all the struct.
Ensure the tags are correct eg no mispell such as `yaml:"key_ids"`

jkongie:
> > > do you want to have a test for loading a yaml file into those struct?
> > 
> > 
> > I wasn't planning to since it's so simple, but I can if you think there is value
> 
> @jkongie we can keep it super simple one test just tries to unmarshal a yaml into all the struct. Ensure the tags are correct eg no mispell such as `yaml:"key_ids"`

Added tests for this

DimitriosNaikopoulos (approved):

PR #304, opened 2025-08-21
refactor(engine): move networks from cld with config injection

ajaskolski:
Moves pkg/environment/networks.go to cld with added config injection.

### DIFF
--- a/engine/cld/network/networks_test.go
+++ b/engine/cld/network/networks_test.go
@@ -1,0 +1,407 @@
+package network
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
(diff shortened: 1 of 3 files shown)


engine/cld/network/networks_test.go
```diff
@@ -0,0 +1,407 @@
...
+		BlockExplorer: cldf_config_network.BlockExplorer{
+			Type:   "Etherscan",
+			APIKey: "test_key",
+			URL:    "https://etherscan.io",
+		},
+		RPCs: []cldf_config_network.RPC{
+			{
+				RPCName:            "test_rpc",
+				PreferredURLScheme: "http",
+				HTTPURL:            "https://gap-rpc-proxy.public.main.prod.cldev.sh:4443/ethereum-mainnet",
```
bytesizedroll, line 250: public, we dont want these here

PR #306, opened 2025-08-21
Update Sui CTF provider configs to allow specifying image and platform

faisal-chainlink:
We sometimes use an `arm64` image instead of the default `amd64` for Sui. This change edits the `CTFChainProviderConfig` for Sui CTF provider to allow passing that config into CTF.


RodrigoAD (approved):
LGTM 👍

graham-chainlink (approved):
LGTM

PR #308, opened 2025-08-21
feat: add more domain paths for the config dirs and files

jkongie:
This change adds more paths for the domain config dirs and files so they can be called on the domain struct. This aims to provide a way to access the expected paths for the domain config dirs and files.

### DIFF
--- a/engine/cld/domain/paths.go
+++ b/engine/cld/domain/paths.go
@@ -1,54 +1,90 @@
 package domain
 
 const (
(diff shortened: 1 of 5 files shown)


engine/cld/domain/paths.go
jkongie, line 1: A bit of clean up was also added to this file to make it more readable, and removed some redundant copy in the comments

graham-chainlink (approved):

PR #312, opened 2025-08-25
fix: correct function names in domain.go and domain_test.go

jkongie:
Fixes to correctly reference the difference between dir and file path.

### DIFF
--- a/engine/cld/domain/domain.go
+++ b/engine/cld/domain/domain.go
@@ -121,13 +121,13 @@ func (d Domain) ConfigLocalDirPath() string {
 	return filepath.Join(d.ConfigDirPath(), DomainConfigLocalDirName)
 }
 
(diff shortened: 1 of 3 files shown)


engine/cld/domain/domain.go
```diff
@@ -122,12 +122,12 @@ func (d Domain) ConfigLocalDirPath() string {
 }
 
 // ConfigLocalFileName returns the path to a domain environment's local execution config file.
```
graham-chainlink, line 124:
```suggestion
// ConfigLocalFilePath returns the path to a domain environment's local execution config file.
```

jkongie: Thanks. Fixed

graham-chainlink (approved):

PR #316, opened 2025-08-25
mover config resolvers to cldf

bytesizedroll:


ajaskolski (commented):
looks good

graham-chainlink:
> I would like to keep the config package related to just setup configuration, rather than the changeset configuration.
> 
> We should discuss where this code should live, but I think there is a better place for it than config.

@jkongie maybe `cld/changeset/configresolver` ?

bytesizedroll:
+1 for @graham-chainlink idea, it is changeset specific as it is generating changeset input so i think that makes sense

jkongie (approved):

PR #327, opened 2025-08-26
feat: adds domain config

giogam:


engine/cld/config/domain/domain.go
```diff
@@ -0,0 +1,63 @@
...
+	v := viper.New()
+	v.SetConfigFile(filePath)
+
+	if err := v.ReadInConfig(); err != nil {
+		return nil, err
+	}
+
+	cfg := &DomainConfig{}
+	err := v.Unmarshal(cfg)
+
```
ajaskolski, line 61: nit. hmm we could return here broken cfg. Might consider verifying it just after marshalling to catch it earlier ?

```
    for name, env := range cfg.Environments {
        if err := env.Validate(); err != nil {
            return nil, fmt.Errorf("invalid environment %q: %w", name, err)
        }
    }
```

engine/cld/config/domain/domain.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/spf13/viper"
+)
+
+// Environment represents a single environment configuration.
+type Environment struct {
+	NetworkAccess []string `mapstructure:"networkAccess" yaml:"networkAccess"`
+}
+
+// Validate validates the environment configuration.
+func (e *Environment) Validate() error {
```
DimitriosNaikopoulos, line 16: Based on this comment https://github.com/smartcontractkit/chainlink-deployments-framework/pull/327/files#r2303218500. It would be better to assume that the `Load` action always returns valid config or error therefore we can make this private since we don't want to force the end user to have to validate the config on their side but just assume that the `Load` function will do it for them

engine/cld/config/domain/domain.go
giogam: Thanks, makes sense!

DimitriosNaikopoulos (dismissed):

engine/cld/config/domain/domain.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/spf13/viper"
+)
+
+// Environment represents a single environment configuration.
+type Environment struct {
+	NetworkAccess []string `mapstructure:"networkAccess" yaml:"networkAccess"`
+}
+
+// validate validates the environment configuration.
+func (e *Environment) validate() error {
```
DimitriosNaikopoulos, line 15: Will this validate only 1 of the networks? For example
```
environments:
  development:
    networkAccess:
      - testnet
  staging:
    networkAccess:
      - testnet
      - invalid
```

Will this function when is called validate access for both `development` and `staging`?

engine/cld/config/domain/domain_test.go
```diff
@@ -0,0 +1,218 @@
...
+				assert.Equal(t, []string{"mainnet"}, testtest.NetworkAccess)
+			},
+		},
+		{
+			name:     "error when file not found",
+			filePath: "/tmp/nonexistent_file.yml",
+			wantErr:  true,
+		},
+		{
+			name:     "invalid yaml content",
```
DimitriosNaikopoulos, line 183: Can we also add a test case with multiple environments that one has an invalid network access? 
eg
```
environments:
  development:
    networkAccess:
      - testnet
  staging:
    networkAccess:
      - testnet
      - invalid
  ```

giogam: yep, I had something similar but I removed it by mistake

ajaskolski (approved):

PR #335, opened 2025-08-27
CLD-522: Migrate Changeset Registry

DimitriosNaikopoulos:
## CLD-522: Migrate Changeset Registry

- Move registry to framework
- Rename instances of `migration` to `changeset`
- Update testing requirements based on new package structure


engine/cld/registry/registry.go
```diff
@@ -0,0 +1,292 @@
...
+func newArchivedRegistryEntry(gitSHA string) registryEntry {
+	return registryEntry{gitSHA: &gitSHA}
+}
+
+// IsArchived returns true if the changeset is archived.
+func (e registryEntry) IsArchived() bool {
+	return e.gitSHA != nil
+}
+
+// ChangesetsRegistry is a registry of changesets that can be applied to a domain environment. Changesets can be either
```
jkongie, line 93: We should be good to get rid of the comments on the differences between active and buried since it's not really applicable to pipelines, and we plan to remove migrations

engine/cld/registry/registry.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	return entry.changeset.Configurations()
+}
+
+// ChangesetOption defines an option for configuring a changeset
+type ChangesetOption func(*ChangesetConfig)
+
+// ChangesetConfig holds configuration options for a changeset
+type ChangesetConfig struct {
```
jkongie, line 168: This could potentially be made private, since it is only used internally

graham-chainlink: I think this is used by other packages as well right? like in the [command package](https://github.com/smartcontractkit/chainlink-deployments/blob/07d456a7d4257b3e54c4fc567eee93e631dfd142/pkg/commands/migration.go#L223-L223)

engine/cld/registry/registry_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+// noopChangesetConfig holds configuration for the noop changeset.
+type noopChangesetConfig struct {
+}
+
+// createNoopChangeset creates a noop changeset for testing.
+func createNoopChangeset(t *testing.T, chainOverrides []uint64) changeset.ChangeSet {
+	t.Helper()
+
+	cs := cldf.CreateChangeSet(
```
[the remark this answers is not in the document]
DimitriosNaikopoulos: We will not have to use this anymore in this test since this will be moved on the `changeset` pkg

engine/cld/registry/registry.go
```diff
@@ -0,0 +1,292 @@
...
+	ChainsToLoad      []uint64
+	WithoutJD         bool
+	OperationRegistry *operations.OperationRegistry
```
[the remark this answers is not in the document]
DimitriosNaikopoulos: Would this still be accessible through the [getter function ](https://github.com/smartcontractkit/chainlink-deployments/blob/07d456a7d4257b3e54c4fc567eee93e631dfd142/pkg/commands/migration_helper.go#L36)?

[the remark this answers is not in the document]
graham-chainlink: The GetMigrationOptions is returning the struct `ChangesetConfig` ,so if we have the fields private, then consumers wont be able to access those fields right?

bytesizedroll (approved):

bytesizedroll (dismissed):

engine/cld/changeset/registry_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	cldf "github.com/smartcontractkit/chainlink-deployments-framework/deployment"
+)
+
+// noopMigration is a migration that does nothing.
+//
+// Used for testing.
+type noopMigration struct {
```
graham-chainlink, line 14: This could be renamed now ?
```suggestion
type noopChangeset struct {
```

PR #336, opened 2025-08-27
feat: adds cognito token source to offchain/jd

jkongie:
This adds functionality to the `offchain/jd` package to create an oauth2 token source for  Cognito. This token is used to authenticate against the Job Distributor API.


offchain/jd/cognito.go
```diff
@@ -0,0 +1,144 @@
...
+func (c *CognitoTokenSource) Authenticate(ctx context.Context) error {
+	// Create client if not already set (for production use)
+	if c.client == nil {
+		sdkConfig, err := config.LoadDefaultConfig(ctx,
+			config.WithRegion(c.auth.AWSRegion),
+		)
+		if err != nil {
+			return err
+		}
+		c.client = cognitoidentityprovider.NewFromConfig(sdkConfig)
```
graham-chainlink, line 69: i wonder should this be done on the `NewCognitoTokenSource` stage so we know early if there is issue with the config? `NewCognitoTokenSource` also supports returning an error whihc we are not using now.

offchain/jd/cognito.go
```diff
@@ -0,0 +1,144 @@
...
+	Password               string
+}
+
+// NewCognitoTokenSource creates a new CognitoTokenSource with the given CognitoAuth configuration.
+// If client is nil, a real AWS Cognito client will be created when Authenticate is called.
+func NewCognitoTokenSource(auth CognitoAuth) (*CognitoTokenSource, error) {
+	return &CognitoTokenSource{
+		auth: auth,
+	}, nil
+}
```
graham-chainlink, line 54: We support returning error here on the 2nd return param but it is not used at this moment. Refer to my other comment maybe we could move the config initialization logic here.

offchain/jd/cognito.go
```diff
@@ -0,0 +1,144 @@
...
+// standard OAuth2 client libraries and HTTP clients that support token sources.
+//
+// Note: This method uses context.Background() for authentication if no cached token exists.
+// For more control over authentication context and timeout behavior, consider calling
+// Authenticate() explicitly before calling Token().
+//
+// Returns an OAuth2 token containing the Cognito access token
+func (c *CognitoTokenSource) Token() (*oauth2.Token, error) {
+	if c.authResult == nil {
+		if err := c.Authenticate(context.Background()); err != nil {
```
graham-chainlink, line 110: Should token accept context too so we dont have to use context.Background() here?

offchain/jd/cognito_test.go
```diff
@@ -0,0 +1,224 @@
...
+	t.Parallel()
+
+	// Test with static values to ensure the hash calculation is correct
+	auth := CognitoAuth{
+		CognitoAppClientID:     "clientid",
+		CognitoAppClientSecret: "secret",
+		Username:               "user",
+	}
+
+	tokenSource, _ := NewCognitoTokenSource(auth)
```
graham-chainlink, line 217: should we handle the error here too?

offchain/jd/cognito_test.go
```diff
@@ -0,0 +1,224 @@
...
+			beforeFunc: func(t *testing.T, client *mocks.MockCognitoClient, source *CognitoTokenSource) {
+				t.Helper()
+
+				// Pre-populate auth result to simulate cached token
+				accessToken := "cached-access-token"
+				source.authResult = &types.AuthenticationResultType{
+					AccessToken: &accessToken,
+				}
+
+				client.AssertNotCalled(t, "InitiateAuth")
```
graham-chainlink, line 157: Should this `AssertNotCalled` be checked later since during the `beforeFunc` we have not called `source.Token()` yet, so this will always be true

offchain/jd/cognito.go
jkongie: Token can't accept a context because the oath2.TokenSource interface does not define any arguments.

The docs explain here that if they want fine grain control, they can call `Authenticate` with a context before calling `Token`

offchain/jd/cognito.go
jkongie: The original implementation had this one the constructor, but I moved all the setup out into an Authenticate function.

The user has two options here:

1. Very simple with little control 

```
s := NewCognitoTokenSource(auth)
s.Token()
```

This will handle everything for them under the hood, but they do not get to fail early when requesting the token.

2. Fine grained control

```
s := NewCognitoTokenSource(auth)
err := s.Authenticate(ctx) // Authenticate early to capture any errors

// .... do whatever they want

s.Token()
```

Coupling it into the constructor means that you must now always pass a context to the constructor which I was trying to avoid.

offchain/jd/cognito.go
jkongie: Updated to remove the error

jkongie: See above for explanation on why I think it should not be in the constructor

offchain/jd/cognito_test.go
jkongie: Removed the error return so this is not needed

offchain/jd/cognito_test.go
jkongie: Yeah you are right. 

I've actually just removed the assert. Since I'm not declaring any expectations, the test will fail if the `InitiateAuth` method is called, so the check for the number of calls is redundant.

graham-chainlink (approved):

graham-chainlink (approved):

PR #337, opened 2025-08-27
chore: refactor camelCase with snake_case

giogam:
Minor refactor replacing `camelCase` with `snake_case` to keep the domain configuration consistent with the rest of the codebase.

### DIFF
--- a/.changeset/three-cherries-deny.md
+++ b/.changeset/three-cherries-deny.md
@@ -1,0 +1,5 @@
+---
+"chainlink-deployments-framework": patch
+---
+
+chore: refactor camelCase with snake_case


DimitriosNaikopoulos (approved):

engine/cld/config/domain/testdata/mixed.yaml
```diff
@@ -1,11 +1,11 @@
 environments:
   dev:
-    networkAccess:
+    network_types:
```
DimitriosNaikopoulos, line 3: I preferred the `network_access` but that is a nit (because is more like network ACLs)

PR #345, opened 2025-08-29
[CLD-570]: fix(JD): migrate dry run client to CLDF

graham-chainlink:
MIgrating dry run JD client from CLD to CLDF as this is needed for loadning environment.

Also introduced ability to create dry run client with the JD provider

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-570

### DIFF
--- a/offchain/jd/dry_run_client.go
+++ b/offchain/jd/dry_run_client.go
@@ -1,0 +1,165 @@
+package jd
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+
+	cldf_offchain "github.com/smartcontractkit/chainlink-deployments-framework/offchain"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	csav1 "github.com/smartcontractkit/chainlink-protos/job-distributor/v1/csa"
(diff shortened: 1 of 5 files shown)


offchain/jd/dry_run_client.go
```diff
@@ -0,0 +1,165 @@
...
+	return &DryRunJobDistributor{
+		realBackend: realBackend,
+		lggr:        lggr,
+	}
+}
+
+// GetJob retrieves a specific job by its ID from the Job Distributor.
+// This operation is forwarded to the real backend since it's a read-only operation.
+func (d *DryRunJobDistributor) GetJob(ctx context.Context, in *jobv1.GetJobRequest, opts ...grpc.CallOption) (*jobv1.GetJobResponse, error) {
+	d.lggr.Infow("DryRunJobDistributor.GetJob", "in", in)
```
jkongie, line 37: I wonder how much value we get out of these loggers. 

If we omitted them, we could just embed `offchain.Client` and not have to implement any of the read methods.

Not a blocker, just a thought

graham-chainlink: Yeah i had the same thought too, but i think we can revisit it since the changes are already there and we already import the logger in CLDF

PR #346, opened 2025-08-29
feat: refresh cognito token when expired

jkongie:
This change inspects the token expiry time each time `Token()` is called on the `CognitoTokenSource` and refreshes the token when it is expired. If the refresh fails, it falls back to full re-authentication.

### DIFF
--- a/offchain/jd/cognito.go
+++ b/offchain/jd/cognito.go
@@ -5,6 +5,7 @@ import (
 	"crypto/hmac"
 	"crypto/sha256"
 	"encoding/base64"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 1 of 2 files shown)


offchain/jd/cognito.go
```diff
@@ -117,6 +136,49 @@ func (c *CognitoTokenSource) Token() (*oauth2.Token, error) {
...
+// without reusing the user's credentials. This is more efficient than full re-authentication
+// and follows OAuth2 best practices.
+//
+// The method:
+//  1. Uses the cached refresh token to call InitiateAuth with REFRESH_TOKEN_AUTH flow
+//  2. Updates the cached authentication result with new tokens
+//  3. Calculates and stores the new token expiry time
+//
+// Returns an error if the refresh fails (e.g., refresh token expired or invalid).
+func (c *CognitoTokenSource) RefreshToken(ctx context.Context) error {
```
graham-chainlink, line 151: Can this be private?

graham-chainlink (approved):

PR #350, opened 2025-09-01
fix: migrate proposejob & registernode

graham-chainlink:
Migrate Propose Job and Register Node from pkg/offchain in CLD to CLDF.

Also added tests as it didn have any.
In order to mock the JD client in engine, i needed to move the mocks in offchain package to a place that is accessible for both offchain and engine package.

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-568


internal/pointer/pointer.go
```diff
@@ -0,0 +1,14 @@
...
+}
+
+func DerefOrEmpty[T any](ptr *T) T {
+	if ptr == nil {
+		var v T
+		return v
+	}
+
+	return *ptr
+}
```
jkongie, line 14: I think this is only used in mcms so we should remove it

graham-chainlink: removed it!

jkongie (approved):

PR #358, opened 2025-09-02
[CLD-586]: feat(engine): migrate loadCatalogStore & loadOffchainClient

graham-chainlink:
In order to load environment, we need to load the catalog store first and also the offchain client. These will be used when we load environment.

- added tests

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-586

### DIFF
--- a/engine/cld/offchain/offchain.go
+++ b/engine/cld/offchain/offchain.go
@@ -1,0 +1,84 @@
+package offchain
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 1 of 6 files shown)


engine/cld/offchain/offchain.go
```diff
@@ -0,0 +1,84 @@
+package offchain
```
graham-chainlink, line 1: Unable to add test for this due to hard dependency on `NewCognitoTokenSource` and `NewClientOffchainProvider` , will have to come back and refactor this.

PR #365, opened 2025-09-03
chore: duplicates multiclient and rpc_config in chain/evm

giogam:
his PR relocates `multiclient.go`, `multiclient_test.go`, `rpc_config.go`, and `rpc_config_test.go` from the `deployment` package to the `evm` package. The files are initially duplicated in both packages; the originals will be removed after all CLD and CLDF references are updated.


deployment/multiclient.go
```diff
@@ -1,3 +1,6 @@
+// Package deployment contains deprecated multiclient functionality.
+// Deprecated: Use github.com/smartcontractkit/chainlink-deployments-framework/chain/evm instead.
+// All types and functions in this file have been moved to chain/evm/multiclient.go.
 package deployment
```
graham-chainlink, line 4: I think you can use type alias here, so if chainlink is updated, and CLD is not, at least it will compile.

chain/evm/multiclient.go
graham-chainlink, line 1: I wonder should we move the multiclient stuff under its own package? What do you think?

giogam: Treating the client as a provider component, it makes sense to place it in the provider package, following the same pattern as the confirm functor.

giogam: I moved it to `provider/rpcclient` to match the structure used by `ton` and `solana`. I think it's a good option as it maintains consistency across providers, let me know your thoughts.

chain/evm/provider/rpcclient/rpc_config.go
```diff
@@ -0,0 +1,67 @@
...
+
+type URLSchemePreference int
+
+const (
+	URLSchemePreferenceNone URLSchemePreference = iota
+	URLSchemePreferenceWS
+	URLSchemePreferenceHTTP
+)
+
+func URLSchemePreferenceFromString(s string) (URLSchemePreference, error) {
```
graham-chainlink, line 17: similar here too , lets add some godoc/commens on public methods and structs

deployment/multiclient.go
```diff
@@ -1,472 +1,17 @@
...
-		DialAttempts: RPCDefaultDialRetryAttempts,
-		DialDelay:    RPCDefaultDialRetryDelay,
-		DialTimeout:  RPCDefaultDialTimeout,
-	}
-}
-
-// MultiClient should comply with the OnchainClient interface
-var _ cldf_evm.OnchainClient = &MultiClient{}
-
-type MultiClient struct {
```
graham-chainlink, line 64: You may want to type alias `MultiClient` too as it is public

giogam: This is done here right? https://github.com/smartcontractkit/chainlink-deployments-framework/blob/843b0e6f86b67237b31305fab44184a43a790d87/deployment/multiclient.go#L10

graham-chainlink (approved):

DimitriosNaikopoulos (approved):

PR #367, opened 2025-09-03
chore: migrate state commands to cld engine as legacy

jkongie:
This commit migrates the state commands to the cld engine as legacy commands.


DimitriosNaikopoulos (commented):
Looks good just needs a changeset to be added

graham-chainlink (approved):

PR #369, opened 2025-09-04
[CLD-528]: feat(command): migrate env command

graham-chainlink:
Migrate env cli commands from CLD

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-528

### DIFF
--- a/engine/cld/legacy/cli/commands/env_test.go
+++ b/engine/cld/legacy/cli/commands/env_test.go
@@ -1,0 +1,53 @@
+package commands
+
+import (
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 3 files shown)


engine/cld/legacy/cli/commands/env_test.go
```diff
@@ -0,0 +1,75 @@
...
+	for _, c := range cmd.Commands() {
+		if c.Name() == name {
+			return c
+		}
+	}
+
+	return nil
+}
+
+func TestNewEnvCmds_BasicStructure(t *testing.T) {
```
DimitriosNaikopoulos, line 21: Should the test use testify instead?

graham-chainlink: cool, updated, i was just copying it from CLD . But now updated to use testify

PR #373, opened 2025-09-04
NONEVM-2498 - Add TON support in CLD

nicolasgnr:
NONEVM-2498 - Add TON support in CLD

### DIFF
--- a/chain/ton/provider/rpc_provider.go
+++ b/chain/ton/provider/rpc_provider.go
@@ -1,0 +1,160 @@
+package provider
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/xssnick/tonutils-go/liteclient"
+	tonlib "github.com/xssnick/tonutils-go/ton"
+	"github.com/xssnick/tonutils-go/ton/wallet"
+	"log"
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain"
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/ton"
+)
+
+// RPCChainProviderConfig holds the configuration to initialize the RPCChainProvider.
+type RPCChainProviderConfig struct {
+	// Required: The HTTP RPC URL to connect to the Ton node.
+	HTTPURL string
+	// Optional: The WebSocket URL to connect to the Ton node.
+	WSURL             string
+	DeployerSignerGen PrivateKeyGenerator
+	WalletVersion     string
+}
+
+// validate checks if the RPCChainProviderConfig is valid.
+func (c RPCChainProviderConfig) validate() error {
+	if c.HTTPURL == "" {
+		return errors.New("rpc url is required")
+	}
+	if c.DeployerSignerGen == nil {
+		return errors.New("deployer signer generator is required")
+	}
+	if _, err := getWalletVersionConfig(c.WalletVersion); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+var _ chain.Provider = (*RPCChainProvider)(nil)
+
+// RPCChainProvider is a chain provider that provides a chain that connects to an TON node via
+// RPC.
+type RPCChainProvider struct {
+	// Ton chain selector, used to identify the chain.
+	selector uint64
+
+	// RPCChainProviderConfig holds the configuration for the RPCChainProvider.
+	config RPCChainProviderConfig
+
+	// chain is the Ton chain instance that this provider manages. The Initialize method
+	// sets up the chain.
+	chain *ton.Chain
+}
+
+func NewRPCChainProvider(selector uint64, config RPCChainProviderConfig) *RPCChainProvider {
+	return &RPCChainProvider{
+		selector: selector,
+		config:   config,
+	}
+}
+
+// Initialize initializes the RPCChainProvider.
+func (p *RPCChainProvider) Initialize(_ context.Context) (chain.BlockChain, error) {
+	if p.chain != nil {
+		return *p.chain, nil // Already initialized
+	}
+
+	// Validate the provider configuration
+	if err := p.config.validate(); err != nil {
+		return nil, fmt.Errorf("failed to validate provider config: %w", err)
+	}
+
+	// Initialize TON client
+	connectionPool := liteclient.NewConnectionPool()
+	//  "https://ton-blockchain.github.io/testnet-global.config.json"
+	// Connect to public LiteServer config. We use the RPC URL to
+	// TODO Ask Blockchain connectivity how to get the correct config url for different networks
+	err := connectionPool.AddConnectionsFromConfigUrl(context.Background(), p.config.HTTPURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve ton network config: %w", err)
+	}
+
+	api := tonlib.NewAPIClient(connectionPool, tonlib.ProofCheckPolicySecure)
+
+	// Wallet
+	privateKey, err := p.config.DeployerSignerGen.Generate()
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate private key: %w", err)
+	}
+
+	// (No need to validate that the version is supported, already done by p.config.validate)
+	walletConfig, _ := getWalletVersionConfig(p.config.WalletVersion)
+	tonWallet, err := wallet.FromPrivateKeyWithOptions(api, privateKey, walletConfig, wallet.WithWorkchain(0))
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to init TON wallet: %w", err)
+	}
+
+	log.Printf("TON wallet loaded with address %s", tonWallet.WalletAddress().String())
+
+	// Create the TON chain instance with the provided configuration
+	p.chain = &ton.Chain{
+		ChainMetadata: ton.ChainMetadata{
+			Selector: p.selector,
+		},
+		Client:        api,
+		Wallet:        tonWallet,
+		WalletAddress: tonWallet.WalletAddress(),
+		URL:           p.config.HTTPURL,
+	}
+
+	return *p.chain, nil
+}
+
+func getWalletVersionConfig(version string) (wallet.VersionConfig, error) {
+	switch version {
+	case "V1R1":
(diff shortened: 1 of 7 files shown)


chain/ton/provider/rpc_provider.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	if err := p.config.validate(); err != nil {
+		return nil, fmt.Errorf("failed to validate provider config: %w", err)
+	}
+
+	// Initialize TON client
+	connectionPool := liteclient.NewConnectionPool()
+	// Connect to public LiteServer config. We use the RPC URL to get the config
+	err := connectionPool.AddConnectionsFromConfigUrl(context.Background(), p.config.HTTPURL)
```
jkongie, line 78: Is there a reason why this can't use the context provided to the `Initialize` method?

Using the provided context means the caller can control any cancellations

chain/ton/provider/rpc_provider.go
```diff
@@ -0,0 +1,158 @@
...
+)
+
+// RPCChainProviderConfig holds the configuration to initialize the RPCChainProvider.
+type RPCChainProviderConfig struct {
+	// Required: The HTTP RPC URL to connect to the Ton node.
+	HTTPURL string
+	// Optional: The WebSocket URL to connect to the Ton node.
+	WSURL             string
+	DeployerSignerGen PrivateKeyGenerator
+	WalletVersion     string
```
jkongie, line 24: I see further down that it is a very specific string that needs to be provided. We should add a comment here explaining what the WalletVersion refers to.

Potentially you may want to also change this type from a general string to a typed string.

chain/ton/provider/rpc_provider.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/ton"
+)
+
+// RPCChainProviderConfig holds the configuration to initialize the RPCChainProvider.
+type RPCChainProviderConfig struct {
+	// Required: The HTTP RPC URL to connect to the Ton node.
+	HTTPURL string
+	// Optional: The WebSocket URL to connect to the Ton node.
+	WSURL             string
+	DeployerSignerGen PrivateKeyGenerator
```
jkongie, line 23: By convention, we always document all the fields in the config so that the caller can easily see what each of the config options are for, and whether they are required.

chain/ton/provider/rpc_provider.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// (No need to validate that the version is supported, already done by p.config.validate)
+	walletConfig, _ := getWalletVersionConfig(p.config.WalletVersion)
+	tonWallet, err := wallet.FromPrivateKeyWithOptions(api, privateKey, walletConfig, wallet.WithWorkchain(0))
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to init TON wallet: %w", err)
+	}
+
+	log.Printf("TON wallet loaded with address %s", tonWallet.WalletAddress().String())
```
jkongie, line 99: We should avoid logging in cases unless absolutely necessary. Logs should generally be implemented by the caller, and not the framework

engine/cld/config/env/config.go
```diff
@@ -208,6 +218,8 @@ var (
 		"onchain.aptos.deployer_key":                              {"ONCHAIN_APTOS_DEPLOYER_KEY", "APTOS_DEPLOYER_KEY"},
 		"onchain.tron.deployer_key":                               {"ONCHAIN_TRON_DEPLOYER_KEY", "TRON_DEPLOYER_KEY"},
 		"onchain.sui.deployer_key":                                {"ONCHAIN_SUI_DEPLOYER_KEY"},
+		"onchain.ton.deployer_key":                                {"ONCHAIN_TON_DEPLOYER_KEY", "TON_DEPLOYER_KEY"},
+		"onchain.ton.wallet_version":                              {"ONCHAIN_TON_WALLET_VERSION", "TON_WALLET_VERSION"},
```
jkongie, line 222: The second string in the slice is the alternative env var key to support backwards compatibility. Since this is a new chain, I would suggest removing it. This is similar to the SUI entry slightly above

```suggestion
		"onchain.ton.deployer_key":                                {"ONCHAIN_TON_DEPLOYER_KEY"},
		"onchain.ton.wallet_version":                              {"ONCHAIN_TON_WALLET_VERSION"},
```

engine/cld/config/env/config_test.go
```diff
@@ -96,6 +102,8 @@ var (
 		"SOLANA_PROGRAM_PATH":               "/tmp",
 		"APTOS_DEPLOYER_KEY":                "0x123",
 		"TRON_DEPLOYER_KEY":                 "0x123",
+		"TON_DEPLOYER_KEY":                  "0x123",
+		"TON_WALLET_VERSION":                "V5R1",
```
jkongie, line 106: You can remove this when you remove the alternative env var key suggested above

chain/ton/provider/rpc_provider.go
nicolasgnr: Documentation added

chain/ton/provider/rpc_provider.go
nicolasgnr: No reason. I've just updated it.

engine/cld/config/env/config.go
```diff
@@ -41,6 +41,15 @@ type SolanaConfig struct {
 	ProgramsDirPath string `mapstructure:"programs_dir_path" yaml:"programs_dir_path"` // The path to the Solana programs directory.
 }
 
+// TonsConfig is the configuration for the Ton Chains.
```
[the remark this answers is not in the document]
nicolasgnr: Suggestion applied in the latest commit

engine/cld/config/env/config.go
nicolasgnr: Suggestion applied in the latest commit

engine/cld/chains/chains_test.go
```diff
@@ -861,3 +879,24 @@ func newFakeRPCServer(t *testing.T) *httptest.Server {
 
 	return srv
 }
+
+// newFakeTONRPCConfigServer returns TON server config server which always answers with a valid
+//
+// When the test is done, the server is closed automatically.
+func newFakeTONRPCConfigServer(t *testing.T) *httptest.Server {
```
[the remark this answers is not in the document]
nicolasgnr: I combined both in a single mock server

chain/ton/provider/rpc_provider.go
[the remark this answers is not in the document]
nicolasgnr: No reason. James made the same suggestion and I updated the code.

DimitriosNaikopoulos (dismissed):

chain/ton/provider/rpc_provider.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	return nil
+}
+
+var _ chain.Provider = (*RPCChainProvider)(nil)
+
+// RPCChainProvider is a chain provider that provides a chain that connects to an TON node via
+// RPC.
+type RPCChainProvider struct {
```
krebernisak, line 66: No need for this type to be public.

The `NewRPCChainProvider` constructor should return the public interface type `chain.Provider`.

chain/ton/provider/rpc_provider.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain"
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/ton"
+)
+
+type WalletVersion string
+
+// Allowed TON wallet versions
+const (
+	WalletVersionV1R1    WalletVersion = "V1R1"
```
krebernisak, line 20: Why not reuse the wallet `Version` type from the lib: https://github.com/xssnick/tonutils-go/blob/master/ton/wallet/wallet.go#L26?

chain/ton/provider/rpc_provider.go
```diff
@@ -0,0 +1,176 @@
...
+
+// RPCChainProviderConfig holds the configuration to initialize the RPCChainProvider.
+type RPCChainProviderConfig struct {
+	// Required: The HTTP RPC URL to connect to the Ton node.
+	HTTPURL string
+	// Optional: The WebSocket URL to connect to the Ton node.
+	WSURL string
+	// Required: A generator for the deployer key. Use PrivateKeyFromRaw to create a deployer
+	// key from a private key.
+	DeployerSignerGen PrivateKeyGenerator
```
krebernisak, line 41: Let's avoid passing on private keys anywhere and use the `Signer` interface.

Private keys should be locked down behind the singer (in one place, locked down, no leaks) - one can get a public key and a signer. The wallet can be created from a pubkey/signer: https://github.com/xssnick/tonutils-go/blob/master/ton/wallet/wallet.go#L177

chain/ton/provider/rpc_provider.go
nicolasgnr: This follows the consistency of [https://github.com/smartcontractkit/chainlink-ton-faucets](https://github.com/smartcontractkit/chainlink-ton-faucets)
 and I don’t want to expose all possible versions for CLD. I’ve updated the list to support only the minimum required versions

chain/ton/provider/rpc_provider.go
nicolasgnr: This is following consistency with the other non-EVM chains (Solana, Aptos, SUI and Tron).

chain/ton/provider/rpc_provider.go
nicolasgnr: This is following consistency with the other non-EVM chains (Solana, Aptos, SUI and Tron) in this repo.

chain/ton/provider/rpc_provider.go
krebernisak: Yes, noticed this after my comment. This is a risky approach. I guess we continue like that for now.

@jkongie why do we have PKs floating around? How do we coordinate to fix this?

jkongie (approved):

PR #377, opened 2025-09-05
chore(engine): standarize import aliases

ajaskolski:


engine/cld/chains/chains.go
```diff
@@ -11,18 +11,18 @@ import (
...
-	cldf_evm_rpcclient "github.com/smartcontractkit/chainlink-deployments-framework/chain/evm/provider/rpcclient"
-	cldf_solana_provider "github.com/smartcontractkit/chainlink-deployments-framework/chain/solana/provider"
-	cldf_sui_provider "github.com/smartcontractkit/chainlink-deployments-framework/chain/sui/provider"
-	cldf_ton_provider "github.com/smartcontractkit/chainlink-deployments-framework/chain/ton/provider"
-	cldf_tron_provider "github.com/smartcontractkit/chainlink-deployments-framework/chain/tron/provider"
+	fchain "github.com/smartcontractkit/chainlink-deployments-framework/chain"
+	aptosprov "github.com/smartcontractkit/chainlink-deployments-framework/chain/aptos/provider"
+	fevm "github.com/smartcontractkit/chainlink-deployments-framework/chain/evm"
+	evmprov "github.com/smartcontractkit/chainlink-deployments-framework/chain/evm/provider"
+	fclient "github.com/smartcontractkit/chainlink-deployments-framework/chain/evm/provider/rpcclient"
```
DimitriosNaikopoulos, line 18: I would have that as `evmclient` because is sub module of `evmprov`

engine/cld/changeset/common.go
```diff
@@ -6,8 +6,8 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/smartcontractkit/chainlink-deployments-framework/changeset/resolvers"
-	"github.com/smartcontractkit/chainlink-deployments-framework/deployment"
+	fresolvers "github.com/smartcontractkit/chainlink-deployments-framework/changeset/resolvers"
```
DimitriosNaikopoulos, line 9: Should we make changes here since there is no conflict for either of these?

engine/cld/domain/fixture_test.go
```diff
@@ -8,8 +8,8 @@ import (
 	"github.com/Masterminds/semver/v3"
 	"github.com/stretchr/testify/require"
 
-	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
-	cldf "github.com/smartcontractkit/chainlink-deployments-framework/deployment"
+	fdatastore "github.com/smartcontractkit/chainlink-deployments-framework/datastore"
```
DimitriosNaikopoulos, line 11: Same as the `fdeployment` should we change it when there is no conflict?

engine/cld/offchain/update_node_test.go
```diff
@@ -12,9 +12,9 @@ import (
 	nodev1 "github.com/smartcontractkit/chainlink-protos/job-distributor/v1/node"
 	"github.com/smartcontractkit/chainlink-protos/job-distributor/v1/shared/ptypes"
 
-	"github.com/smartcontractkit/chainlink-deployments-framework/internal/pointer"
-	"github.com/smartcontractkit/chainlink-deployments-framework/internal/testing/jd/mocks"
-	"github.com/smartcontractkit/chainlink-deployments-framework/offchain/node"
+	fpointer "github.com/smartcontractkit/chainlink-deployments-framework/internal/pointer"
+	jdmocks "github.com/smartcontractkit/chainlink-deployments-framework/internal/testing/jd/mocks"
+	fnode "github.com/smartcontractkit/chainlink-deployments-framework/offchain/node"
```
DimitriosNaikopoulos, line 17: Same as above

DimitriosNaikopoulos (commented):
Do we want to change pkg names to `f...` if there is no conflict? Also do we want to make that change if the packages are imported within the framework (ie outside of `engine/*`)?

engine/cld/changeset/common.go
ajaskolski: just sticks to convention here

PR #378, opened 2025-09-05
chore(engine): add note not to delete environment names constant file

ajaskolski:
Add comment to not delete the file.


DimitriosNaikopoulos (changes requested):
These are used by domains

DimitriosNaikopoulos (approved):

PR #380, opened 2025-09-05
chore(framework): standarize import aliases

ajaskolski:
Cleaning import aliases.

- Removes underscores and framework prefixes within framework code.
- Rule of thumb use max 2 words phrases for imports to be readable.
- Shorten import aliases names where valid.
- For conflicting names for imports you can add `f` as prefix to mitigate issue or just use two words from package path, whatever makes more sense.

### DIFF
--- a/chain/evm/evm_chain.go
+++ b/chain/evm/evm_chain.go
@@ -10,7 +10,7 @@ import (
 	"github.com/zksync-sdk/zksync2-go/accounts"
 	"github.com/zksync-sdk/zksync2-go/clients"
 
-	chain_common "github.com/smartcontractkit/chainlink-deployments-framework/chain/internal/common"
+	chaincommon "github.com/smartcontractkit/chainlink-deployments-framework/chain/internal/common"
 )
 
 // ConfirmFunc is a function that takes a transaction, waits for the transaction to be confirmed,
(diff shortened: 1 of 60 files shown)


chain/evm/evm_chain.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/zksync-sdk/zksync2-go/accounts"
 	"github.com/zksync-sdk/zksync2-go/clients"
 
-	chain_common "github.com/smartcontractkit/chainlink-deployments-framework/chain/internal/common"
+	chaincommon "github.com/smartcontractkit/chainlink-deployments-framework/chain/internal/common"
```
DimitriosNaikopoulos, line 13: Should this be just `common`?

chain/sui/provider/rpc_provider.go
```diff
@@ -5,9 +5,9 @@ import (
 	"errors"
 	"fmt"
 
-	sui_sdk "github.com/block-vision/sui-go-sdk/sui"
+	suisdk "github.com/block-vision/sui-go-sdk/sui"
```
DimitriosNaikopoulos, line 8: Should this be `sui`?

chain/tron/tron_chain.go
```diff
@@ -9,11 +9,11 @@ import (
 	"github.com/fbsobreira/gotron-sdk/pkg/http/soliditynode"
 	"github.com/smartcontractkit/chainlink-tron/relayer/sdk"
 
-	cld_common "github.com/smartcontractkit/chainlink-deployments-framework/chain/internal/common"
+	chaincommon "github.com/smartcontractkit/chainlink-deployments-framework/chain/internal/common"
```
DimitriosNaikopoulos, line 12: Should this be `common`? (same as above)

DimitriosNaikopoulos (commented):
Couple of questions about adding path prefixes to some imported pkgs

chain/tron/tron_chain.go
ajaskolski: same as first comment there

DimitriosNaikopoulos (approved):

PR #391, opened 2025-09-08
TON: connection with liteserver url

jadepark-dev:
This PR updates TON provider's behavior to directly use the `liteserver://` connection string.
This PR also bumps CTF/TON to use the updated MyLocalTON Image([v3.7](https://github.com/neodix42/mylocalton-docker/releases)) in testing environment

Related PRs:
- https://github.com/smartcontractkit/chainlink-ton/pull/155
- https://github.com/smartcontractkit/chainlink-testing-framework/pull/2089

### DIFF
--- a/chain/ton/provider/ctf_provider.go
+++ b/chain/ton/provider/ctf_provider.go
@@ -18,7 +18,6 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/xssnick/tonutils-go/address"
-	"github.com/xssnick/tonutils-go/liteclient"
 	"github.com/xssnick/tonutils-go/tlb"
 	"github.com/xssnick/tonutils-go/ton"
 	"github.com/xssnick/tonutils-go/ton/wallet"
@@ -121,7 +120,7 @@ func (p *CTFChainProvider) startContainer(chainID string) (string, *ton.APIClien
 			Type:    blockchain.TypeTon,
 			ChainID: chainID,
 			Port:    strconv.Itoa(port),
-			Image:   "ghcr.io/neodix42/mylocalton-docker:dev", // as of 11th Aug 2025, the tag 'latest' image used by default fails to startup. Override it with the 'dev' image
+			Image:   "ghcr.io/neodix42/mylocalton-docker:v3.7",
 		})
 		if rerr != nil {
 			// Return the ports to freeport to avoid leaking them during retries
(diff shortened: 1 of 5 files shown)


chain/ton/provider/ctf_provider.go
```diff
@@ -202,21 +197,26 @@ func fundTonWallets(t *testing.T, client ton.APIClientWrapped, recipients []*add
 	// we don't wait for the transaction to be confirmed here, as it may take some time
 }
 
-func checkConnection(t *testing.T, client *ton.APIClient) {
+func checkConnection(t *testing.T, client *ton.APIClient) *ton.BlockIDExt {
```
graham-chainlink, line 200: nitpick: i feel like this is no longer a `checkConnection` , but rather a `getTrustedBlock` , maybe we can rename for its purpose?

chain/ton/provider/rpc_provider.go
```diff
@@ -122,6 +187,38 @@ func (p *RPCChainProvider) Initialize(ctx context.Context) (chain.BlockChain, er
...
+	// Setup connection timeout - mirrors tonutils-go AddConnectionsFromConfig logic
+	timeout := 3 * time.Second
+	if dl, ok := ctx.Deadline(); ok {
+		timeout = time.Until(dl)
+	}
+
+	connCtx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	err := pool.AddConnection(connCtx, hostPort, publicKey)
```
graham-chainlink, line 214: why not just this?
```suggestion
	err := pool.AddConnection(ctx, hostPort, publicKey)
```

if the `ctx` has its deadline, then it will be respected when it is pased into `pool.AddConnection`  instead of manually calculating it ?

chain/ton/provider/rpc_provider_test.go
```diff
@@ -72,135 +72,153 @@ func Test_RPCChainProviderConfig_validate(t *testing.T) {
 	}
 }
 
-func Test_RPCChainProvider_Initialize(t *testing.T) {
+func Test_RPCChainProvider_Initialize_AlreadyInitialized(t *testing.T) {
```
graham-chainlink, line 75:
```suggestion
func Test_RPCChainProvider_Initialize(t *testing.T) {
```
why is this call AlreadyInitialized? Seems to be only initialized once?

chain/ton/provider/ctf_provider.go
jadepark-dev: yeah I can find a better name. It still checks the connection with CTF-created TON network, if there is a problem with the chain, we can't query the masterchain block

graham-chainlink (approved):
LGTM

graham-chainlink (approved):

PR #429, opened 2025-09-15
Cld 630/jira pkg

bytesizedroll:


jira/client.go
```diff
@@ -0,0 +1,90 @@
...
+	baseURL    string
+	username   string
+	token      string
+	httpClient *http.Client
+}
+
+// JiraIssue represents a JIRA issue response
+type JiraIssue struct {
+	Key    string                 `json:"key"`
+	Fields map[string]interface{} `json:"fields"`
```
DimitriosNaikopoulos, line 26: Should this be:


```suggestion
	Fields map[string]any `json:"fields"`
```

jira/client_test.go
```diff
@@ -0,0 +1,326 @@
...
+				t.Errorf("Expected token %s, got %s", tt.token, client.token)
+			}
+
+			if client.httpClient == nil {
+				t.Errorf("Expected httpClient to be set")
+			}
+
+			if client.httpClient.Timeout != 30*time.Second {
+				t.Errorf("Expected timeout 30s, got %v", client.httpClient.Timeout)
+			}
```
DimitriosNaikopoulos, line 91: Should we use `assert` for testing since we are using it for other tests? 

eg `assert.Equal(...)`

jira/client_test.go
```diff
@@ -0,0 +1,326 @@
...
+			if err != nil {
+				t.Errorf("Unexpected error: %v", err)
+
+				return
+			}
+
+			if client == nil {
+				t.Errorf("Expected client but got nil")
+
+				return
```
DimitriosNaikopoulos, line 69: Could use `assert.Nil()`

jira/config.go
DimitriosNaikopoulos, line 1: I wonder if this should be part of the existing config loading pkg https://github.com/smartcontractkit/chainlink-deployments-framework/tree/main/engine/cld/config 🤔 Thoughts?

jira/config.go
```diff
@@ -0,0 +1,130 @@
...
+	domainName := parts[0]
+	domainPath := filepath.Join(domainsPath, domainName)
+
+	// Verify the domain directory exists
+	if _, err := os.Stat(domainPath); err != nil {
+		return "", fmt.Errorf("domain directory does not exist: %s", domainPath)
+	}
+
+	return domainPath, nil
+}
```
DimitriosNaikopoulos, line 130: Should the domain be provided by the caller? Then we will not need to find it based on the current working directory 🤔

jira/config_test.go
DimitriosNaikopoulos, line 1: Same as above test file can use `assert` and `require` for testing

jira/converter.go
```diff
@@ -0,0 +1,55 @@
...
+		return zero, fmt.Errorf("failed to load domain JIRA config: %w", err)
+	}
+
+	// Extract all JIRA field names from the field maps
+	var fieldsToFetch = config.GetJiraFields()
+
+	var domain = strings.ToUpper(config.Domain)
+
+	// 2. Get JIRA token from environment variable
+	token := os.Getenv("JIRA_TOKEN_" + domain)
```
DimitriosNaikopoulos, line 31: I think we don't want to expose this to the framework. Instead the caller should call `NewClient` and provide the client as an input in `JiraToStruct` function

jira/mapping.go
```diff
@@ -0,0 +1,104 @@
...
+	var result T
+
+	if issue == nil {
+		return result, errors.New("nil issue")
+	}
+	if config == nil {
+		return result, errors.New("nil config")
+	}
+
+	remappedData := make(map[string]interface{}, len(config.FieldMaps))
```
DimitriosNaikopoulos, line 21: I think usually the linter flags this as:


```suggestion
	remappedData := make(map[string]any, len(config.FieldMaps))
```

jira/mapping.go
```diff
@@ -0,0 +1,104 @@
...
+	if err := json.Unmarshal(jsonBytes, &result); err != nil {
+		return result, fmt.Errorf("failed to unmarshal to target struct (type mismatch - user specified wrong type): %w", err)
+	}
+
+	return result, nil
+}
+
+// getByPath resolves a dotted path against the Jira issue.
+// If the path doesn't start with "fields." or "key", it's treated as "fields.<path>".
+func getByPath(issue *JiraIssue, path string) (interface{}, bool) {
```
DimitriosNaikopoulos, line 51:
```suggestion
func getByPath(issue *JiraIssue, path string) (any, bool) {
```

jira/mapping.go
```diff
@@ -0,0 +1,104 @@
...
+	if issue == nil || path == "" {
+		return nil, false
+	}
+
+	if !strings.HasPrefix(path, "fields.") && path != "key" && !strings.HasPrefix(path, "key.") {
+		path = "fields." + path
+	}
+
+	root := map[string]interface{}{
```
DimitriosNaikopoulos, line 60:
```suggestion
	root := map[string]any{
```

DimitriosNaikopoulos (commented):
A couple of comments about:
- Where the pkg should live
- Using the same testing libraries as with other tests
- Use of `any` instead of `interface{}` (I thought the linter complains about that 🤔 )

engine/cld/config/config.go
```diff
@@ -23,6 +24,10 @@ type Config struct {
 	// and deployment settings. This configuration varies between environments
 	// (development, staging, production) and contains sensitive data.
 	Env *cfgenv.Config
+
+	// Jira contains JIRA integration configuration including connection details
+	// and field mappings for using JIRA in resolvers.
+	Jira *cfgjira.JiraConfig
```
graham-chainlink, line 30: Following the above pattern `cfgenv.Config`, should  we just call JiraConfig  JIRA, eg `cfgjira.Config` ?
```suggestion
	Jira *cfgjira.Config
```

graham-chainlink (commented):
i think you need to run `go mod tidy` on the root go.mod file

engine/cld/config/domain/domain.go
```diff
@@ -59,6 +62,13 @@ func (cfg *DomainConfig) validate() error {
 		}
 	}
 
+	// Validate JIRA config if present (it's optional)
+	if cfg.Jira != nil {
+		if err := cfg.Jira.Validate(); err != nil {
+			return errors.Join(errors.New("invalid JIRA configuration"), err)
```
jkongie, line 68: The standard way to propagate errors is to wrap the error rather than join. Joining is generally used when you want to group errors at the same level, rather than an error coming up from a call 

```suggestion
			return fmt.Errorf("invalid JIRA configuration: %w", err)
```

engine/cld/config/config.go
```diff
@@ -38,8 +43,16 @@ func Load(dom fdomain.Domain, env string, lggr logger.Logger) (*Config, error) {
 		return nil, fmt.Errorf("failed to load env config: %w", err)
 	}
 
+	jiraCfg, err := LoadJiraConfig(dom)
+	if err != nil {
+		// JIRA config is optional, so we only log the error and continue
+		lggr.Warnf("JIRA config not available: %v", err)
```
jkongie, line 49: Since Jira config is optional, we should use `Info` instead of `Warn`.

Or better yet make a distinction between the jira config not present vs an actual error loading the config.

This would allow you to continue when it is not present, but error when they provide it but it is not valid

engine/cld/config/jira.go
```diff
@@ -0,0 +1,27 @@
...
+
+	// Load the full domain config (this handles validation including JIRA)
+	domainConfig, err := domain.Load(domainConfigPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Extract the JIRA config (validation already done by domain.Load)
+	if domainConfig.Jira == nil {
+		return nil, errors.New("no JIRA configuration found in domain config")
```
jkongie, line 23: If you return a sentinel error here, you can make a decide how you want to handle it at the call site.

`var ErrConfigNotFound = errors.New("no Jira configuration found in domain config")`

jira/client.go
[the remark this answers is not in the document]
jkongie: I think this does belong in the engine since it is the only place it is used.

engine/cld/config/config.go
bytesizedroll: done

PR #437, opened 2025-09-16
refactor!: refactor LoadOffchainClient arguments

jkongie:
[BREAKING CHANGE] Refactors the offchain client arguments to remove unnecessary arguments. Converts some of the arguments into functional options with defaults.

Adds tests for the offchain client.


engine/cld/legacy/cli/commands/jd.go
```diff
@@ -1061,7 +1062,10 @@ func loadOffchainClient(
 		return nil, err
 	}
 
-	return offchain.LoadOffchainClient(ctx, domain, envKey, cfg, lggr, true)
+	return offchain.LoadOffchainClient(ctx, domain, cfg.Offchain.JobDistributor,
```
jkongie, line 1065: We do not allow skipping initializing JD here, otherwise the JD commands wouldn't work

engine/cld/internal/credentials/credentials.go
```diff
@@ -1,4 +1,4 @@
-package internal
+package credentials
```
jkongie, line 1: Fix naming

engine/cld/offchain/offchain.go
```diff
@@ -2,83 +2,195 @@ package offchain
...
-				Password:        auth.Password,
-				AWSRegion:       auth.AWSRegion,
-			})
-			if err := source.Authenticate(ctx); err != nil {
+
+		return nil, nil //nolint:nilnil // We want to return nil if the JD is not initialized for now.
+	}
+
+	if endpoints.WSRPC == "" || endpoints.GRPC == "" {
+		return nil, ErrEndpointsRequired
```
jkongie, line 128: Instead of skipping as previously implemented, we return a sentinel error to let the caller decide how they wish to proceed. There are valid cases where you may want to fail entirely instead of skipping the client load. (See the JD commands case)

engine/cld/offchain/offchain.go
```diff
@@ -2,83 +2,195 @@ package offchain
...
+	}
+
+	lggr.Info("Initializing JD client")
+
+	// Setup the oauth2 token source.
+	var oauth oauth2.TokenSource
+	if auth != nil {
+		if loadCfg.tokenSource != nil {
+			oauth = loadCfg.tokenSource // Used for injecting a mock token source for testing.
+		} else {
```
jkongie, line 138: I don't love this because it modifies the code to support testing, but not sure if there is a better way to handle these in unit tests

engine/cld/environment/fork_test.go
```diff
@@ -88,15 +87,6 @@ func Test_LoadForkedEnvironment(t *testing.T) {
...
 
-			if tt.expectPanic {
-				assert.Panics(t, func() {
-					_, err := LoadFork(t.Context(), tt.domain, tt.env, tt.blockNumbers, tt.options...)
-					require.NoError(t, err)
-				})
-
-				return
-			}
-
```
jkongie, line 99: This was panicking because JD was initialized with nil and accessing it caused a panic.

Since we return an error instead we test for the error instead

.changeset/bright-taxes-stick.md
```diff
@@ -0,0 +1,35 @@
...
+
+**Example:**
+```go
+// Old
+LoadOffchainClient(ctx, domain, "testnet", config, logger, false)
+
+// New
+LoadOffchainClient(ctx, domain, config.Offchain.JobDistributor,
+    WithLogger(logger),
+    WithDryRun(true), // Note: inverted!
```
graham-chainlink, line 34: 👍 nice

.mockery.yml
```diff
@@ -24,14 +24,23 @@ packages:
       filename: "mock_{{.InterfaceName | snakecase}}.go"
     interfaces:
       Client:
-  github.com/smartcontractkit/chainlink-deployments-framework/offchain/jd:
+  github.com/smartcontractkit/chainlink-deployments-framework/offchain:
```
graham-chainlink, line 27: Actually there are JD mocks under `/internal/testing/jd/mocks`, maybe we can reuse some there too?

I had it there because i needed to use the same mock across different package

engine/cld/offchain/offchain.go
```diff
@@ -2,83 +2,195 @@ package offchain
...
+type LoadOffchainClientOpt func(*loadConfig)
+
+// WithDryRun sets the dry run mode for the offchain client.
+//
+// When true, the offchain client will perform read operations but no write operations.
+func WithDryRun(dryRun bool) LoadOffchainClientOpt {
+	return func(c *loadConfig) {
+		c.dryRun = dryRun
+	}
+}
```
graham-chainlink, line 69: i wonder if having the option alone should set it to true instead of accepting a boolean flag? Seems cleaner?

```
WithDryRun() // this means we want dry run
```

we never will use the `WithDryRun(false)` anyway since it is false by default

engine/cld/environment/environment_test.go
```diff
@@ -55,16 +55,24 @@ func Test_Load_LoadNodesFailure(t *testing.T) {
 	assert.Contains(t, err.Error(), "nodes.json: no such file or directory")
 }
 
-func Test_Load_LoadOffchainClientFailure(t *testing.T) {
-	t.Parallel()
+func Test_Load_LoadOffchainClientFailure(t *testing.T) { //nolint:paralleltest // This test is not parallel safe due to setting of env vars
+	// When loading the environment in CI, it checks for the CI flag, and uses env vars to load
+	// the config instead of from the config file. To overcome this, we need to manually set the
```
graham-chainlink, line 60: missing some words?
```suggestion
	// the config instead of ??? from the config file. To overcome this, we need to manually set the
```

engine/cld/offchain/offchain.go
```diff
@@ -2,83 +2,195 @@ package offchain
...
+
+// WithCredentials sets the gRPC transport credentials for the offchain client.
+func WithCredentials(creds credentials.TransportCredentials) LoadOffchainClientOpt {
+	return func(c *loadConfig) {
+		c.creds = creds
+	}
+}
+
+// withOffchainProvider sets the offchain provider for the offchain client. Used for testing.
+func withOffchainProvider(provider foffchain.Provider) LoadOffchainClientOpt {
```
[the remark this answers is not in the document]
graham-chainlink: should this be in the go doc comment itself? Else in future when someone else reads it, they will ask why is this unexported?

engine/cld/offchain/offchain.go
jkongie: I used the boolean flag here because it made the interface cleaner and I didn't have to collect options in it's own slice.

If I removed it the argument, the caller would have to do

```go
opts := []Opt

if x == true {
  opts = append(opts, WithDryRun())
}

```

Instead with the boolean flag they can pass it in directly

```
Load(WithDryRun(x))
```

I think this is slightly cleaner, but can see an argument against it for a cleaner option.

engine/cld/environment/environment_test.go
jkongie: Reading it makes sense to me but I can clean it up

ajaskolski (approved):

bytesizedroll (approved):

PR #439, opened 2025-09-16
[CLD-634]: test: separate unit and provider tests

graham-chainlink:
In order to reduce tests execution time, i broken the original test step into 2, one for normal tests and the other for provider tests (with docker containers). By running both parallely, tests should finish faster.

Originally the test runs for around 5 minutes, now since we broken it down into 2 , the provider tests take 3 min and the non provider tests take 1 min , so max of both of them is 3.5 min, so we shave off ~1.5-2 minutes through my initial observation.

Screenshot 2025-09-16 at 11 39 40 pm

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-634

### DIFF
--- a/.github/workflows/pull-request-main.yml
+++ b/.github/workflows/pull-request-main.yml
@@ -38,20 +38,42 @@ jobs:
       actions: read
     steps:
       - name: Build and test
-        uses: smartcontractkit/.github/actions/ci-test-go@eeb76b5870e3c17856d5a60fd064a053c023b5f5 # ci-test-go@1.0.0
+        uses: smartcontractkit/.github/actions/ci-test-go@dfcba48f05933158428bce867d790e3d5a9baa6b # ci-test-go@1.1.0
+        with:
+          # disable the checkptr runtime check due a false positive in github.com/xssnick/tonutils-go
+          # causing tests in ci to fail "fatal error: checkptr: pointer arithmetic result points to invalid allocation"
+          # https://github.com/xssnick/tonutils-go/issues/310
+          # Exclude provider packages which use Docker containers
+          go-test-cmd: go test -race -gcflags=all=-d=checkptr=0 -coverprofile=coverage.txt $(go list ./... | grep -v '/provider')
+          use-go-cache: true
+          artifact-name: unit-tests
+
+  ci-test-provider:
+    name: Provider Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      id-token: write
+      contents: read
+      actions: read
+    steps:
+      - name: Build and test provider packages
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-main.yml
```diff
@@ -38,20 +38,42 @@ jobs:
...
+      - name: Build and test provider packages
+        uses: smartcontractkit/.github/actions/ci-test-go@dfcba48f05933158428bce867d790e3d5a9baa6b # ci-test-go@1.1.0
         with:
           # disable the checkptr runtime check due a false positive in github.com/xssnick/tonutils-go
           # causing tests in ci to fail "fatal error: checkptr: pointer arithmetic result points to invalid allocation"
           # https://github.com/xssnick/tonutils-go/issues/310
           # -p 2 -parallel 3 = 2 packages, 3 tests max = 6 containers max
-          go-test-cmd: go test -race -gcflags=all=-d=checkptr=0 -p 2 -parallel 3 -coverprofile=coverage.txt $(go list ./...)
+          # Only run provider packages which use Docker containers
+          go-test-cmd: go test -race -gcflags=all=-d=checkptr=0 -p 2 -parallel 3 -coverprofile=coverage.txt $(go list ./... | grep '/provider')
```
jkongie, line 68: Would it be better to use build flags instead of managing this by grepping for a dir name?

.github/workflows/pull-request-main.yml
```diff
@@ -38,20 +38,42 @@ jobs:
...
+        with:
+          # disable the checkptr runtime check due a false positive in github.com/xssnick/tonutils-go
+          # causing tests in ci to fail "fatal error: checkptr: pointer arithmetic result points to invalid allocation"
+          # https://github.com/xssnick/tonutils-go/issues/310
+          # Exclude provider packages which use Docker containers
+          go-test-cmd: go test -race -gcflags=all=-d=checkptr=0 -coverprofile=coverage.txt $(go list ./... | grep -v '/provider')
+          use-go-cache: true
+          artifact-name: unit-tests
+
+  ci-test-provider:
```
jkongie, line 51: Adding a new test here will miss the required check. We will have to update branch protections to ensure this is covered

graham-chainlink: yeah, i am aware of this, once this PR is merged, i will update the github checks

.github/workflows/pull-request-main.yml
graham-chainlink: i tried using build tags but it didn felt better, maybe worst. Because of the build behavior

- `go test -tag=docker ./...`  includes all tests without tags and tests with docker tag
- `go test ./...` includes all tests without tags.

Do you see the overlap? Because the go build will include files without tags and not the opposite which is want we want here. So we cant just use build tags. Where as the current approach i have in the PR is simple enough to maintain for now.

jkongie (approved):

PR #500, opened 2025-10-08
[CLD-672]: feat(engine/test): support memory catalog

graham-chainlink:
Introduce `WithCatalog` option for Load environment to enable catalog. I decided to have catalog disable by default because enabling catalog will spin up a new postgres and increase loading time.

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-672


engine/test/environment/components.go
```diff
@@ -12,8 +12,9 @@ import (
 type components struct {
 	mu sync.Mutex
 
-	Chains []fchain.BlockChain
-	Logger logger.Logger
+	Chains      []fchain.BlockChain
+	Logger      logger.Logger
+	withCatalog bool
```
jkongie, line 17: `catalogEnabled` would be a better name here

PR #910, opened 2026-04-03
[CLD-1777]: feat(operations-gen): add evm support

graham-chainlink:
Add evm support for the operations-gen cli, ported the code from https://github.com/smartcontractkit/chainlink-ccip/blob/main/chains/evm/cmd/operations-gen/main.go, improved and adapted it to the new design

I have follow up PR adding some golden tests - https://github.com/smartcontractkit/chainlink-deployments-framework/pull/920

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-1777

### DIFF
--- a/tools/operations-gen/README.md
+++ b/tools/operations-gen/README.md
@@ -1,0 +1,161 @@
+# operations-gen
+
+Generates type-safe Go operation wrappers for smart contracts from their ABIs.
+
+## Usage
+
+```bash
+go run ./tools/operations-gen -config /path/to/operations_gen_config.yaml
+```
+
+The `-config` path can be absolute or relative to the current working directory.
+
+## Project structure
+
+```text
+tools/operations-gen/
+  main.go                         # CLI entrypoint + chain-family dispatch
+  templates/
+    evm/
+      operations.tmpl             # EVM codegen template
+  internal/
+    core/
+      core.go                     # Shared config + helpers/interfaces
+    families/
+      evm/
+        evm.go                    # EVM handler implementation
+        evm_test.go               # EVM unit tests
+        evm_golden_test.go        # End-to-end golden generation tests
+  testdata/
+    evm/                          # ABI/bytecode/config/golden fixtures
+```
+
+`main.go` intentionally stays thin: it parses top-level config, loads the template for the selected chain family, and dispatches to the family handler. Shared helpers and common config types live in `internal/core`.
+
+## Configuration
+
+Create an `operations_gen_config.yaml` alongside your ABI/bytecode directories:
+
+```yaml
+version: "1.0.0"
+chain_family: evm # Optional: defaults to "evm"
+
+input:
+  abi_base_path: "./abi" # Directory containing versioned ABI json files
+  bytecode_base_path: "./bytecode" # Directory containing versioned bytecode .bin files
+
+output:
+  base_path: "." # Directory where generated operations/ folders are written
+
+contracts:
+  - contract_name: FeeQuoter
+    version: "1.6.0"
+    package_name: fee_quoter # Optional: override default package name
+    abi_file: "fee_quoter.json" # Optional: override default ABI filename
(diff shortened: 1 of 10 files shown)


tools/operations-gen/evm.go
```diff
@@ -0,0 +1,877 @@
...
+	// anyType is the fallback Go type for unknown source types.
+	anyType = "any"
+)
+
+// nameOverrides provides special-case naming for specific EVM contracts
+// where the default snake_case conversion produces unexpected results.
+var nameOverrides = map[string]string{
+	"OnRamp":  "onramp",
+	"OffRamp": "offramp",
+}
```
jkongie: I don't think we actually want to support this product specific overrides. Can ythey do it in configuration instead?

jkongie (commented):
We should add more documentation to this

tools/operations-gen/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+output:
+  base_path: "." # Directory where generated operations/ folders are written
+
+contracts:
+  - contract_name: FeeQuoter
+    version: "1.6.0"
+    package_name: fee_quoter # Optional: override default package name
+    abi_file: "fee_quoter.json" # Optional: override default ABI filename
+    no_deployment: false # Optional: skip bytecode and Deploy operation
```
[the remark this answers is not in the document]
graham-chainlink: but if no_deployment is false, in order to generate the Deploy code eg 
```go
var Deploy = contract.NewDeploy(contract.DeployParams[ConstructorArgs]{
	Name:        "link-token:deploy",
	Version:     Version,
	Description: "Deploys the LinkToken contract",
	ContractMetadata: &bind.MetaData{
		ABI: LinkTokenABI,
		Bin: LinkTokenBin,
	},
	BytecodeByTypeAndVersion: map[string]contract.Bytecode{
		cldf_deployment.NewTypeAndVersion(ContractType, *Version).String(): {
			EVM: common.FromHex(LinkTokenBin),
		},
	},
	Validate: func(ConstructorArgs) error { return nil },
})
```

If requires the Bin `LinkTokenBin` ,and `LinkTokenBin` is only used for this purpose, so if `no_deployment` is true, we dont need the `LinkTokenBin`, thats why they are coupled. So i was thinking to keep it as single flag instead?

I will rename it to omit_deploy

tools/operations-gen/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+```yaml
+version: "1.0.0"
+chain_family: evm # Optional: defaults to "evm"
+
+input:
+  base_path: "." # Directory containing abi/ and bytecode/ subdirectories
+
+output:
+  base_path: "." # Directory where generated operations/ folders are written
```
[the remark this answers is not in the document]
graham-chainlink: Cool, not a big lift, i can do that for the extra flexibility

tools/operations-gen/testdata/evm/operations_gen_config.yaml
```diff
@@ -13,7 +14,7 @@ contracts:
     abi_file: "link_token.json"
     functions:
       - name: transfer
-        access: owner
+        access: public
```
graham-chainlink, line 17: Seems like deploy link token ABI doesnt support owner method , so changing this to public

tools/operations-gen/templates/evm/operations.tmpl
```diff
@@ -74,19 +74,19 @@ func (c *{{$.ContractType}}Contract) {{.Name}}({{.Params}}) {{.Returns}} {
 
 type {{.Name}} struct {
 {{- range .Fields}}
-	{{.GoName}} {{.GoType}}
+	{{.GoName}} {{.GoType}} `json:"{{.JSONTag}}"`
```
graham-chainlink: abi field name can be `_to` which is not exported in go, so we want to differentiate between go and json field name.

jkongie (approved):

PR #924, opened 2026-04-11
feat: populate Config attribute in hook params

gustavogama-cll:
This pull request adapts the main hook setup logic in the CLD engine so that the `Config` field from the `PreHookParams` and `PostHookParams` types (which are passed to hook functions) is properly populated.

Additionally, it adds a homonym field to the `PostProposalHookParams` type and deprecates the existing `Input` field, thus ensuring that all hook types provide a consistent interface.

To populate the field with the same value that is passed to changesets, a new `resolvedInput(input string)` method was added to the `internalChangeset` interface. This method simply returns the output of the changeset's configuration providers.

---
CLD-1819

### DIFF
--- a/engine/cld/changeset/common.go
+++ b/engine/cld/changeset/common.go
@@ -296,10 +296,6 @@ func (ccs ChangeSetImpl[C]) Apply(env fdeployment.Environment) (fdeployment.Chan
 }
 
 func (ccs ChangeSetImpl[C]) applyWithInput(env fdeployment.Environment, input any) (fdeployment.ChangesetOutput, error) {
-	if input == nil || reflect.ValueOf(input).IsZero() {
-		return ccs.Apply(env)
-	}
-
 	cInput, ok := input.(C)
 	if !ok {
 		return fdeployment.ChangesetOutput{}, fmt.Errorf("invalid input type: expected %T but got %T", *new(C), input)
--- a/engine/cld/changeset/common.go
+++ b/engine/cld/changeset/common.go
@@ -296,7 +296,7 @@ func (ccs ChangeSetImpl[C]) Apply(env fdeployment.Environment) (fdeployment.Chan
 }
 
 func (ccs ChangeSetImpl[C]) applyWithInput(env fdeployment.Environment, input any) (fdeployment.ChangesetOutput, error) {
-	if reflect.ValueOf(input).IsZero() {
+	if input == nil || reflect.ValueOf(input).IsZero() {
 		return ccs.Apply(env)
 	}
 
--- a/engine/cld/changeset/common.go
+++ b/engine/cld/changeset/common.go
@@ -32,8 +32,9 @@ type Configurations struct {
 type internalChangeSet interface {
 	noop() // unexported function to prevent arbitrary structs from implementing ChangeSet.
 	Apply(env fdeployment.Environment) (fdeployment.ChangesetOutput, error)
-	applyWithInput(env fdeployment.Environment, inputStr string) (fdeployment.ChangesetOutput, error)
 	Configurations() (Configurations, error)
+	applyWithInput(env fdeployment.Environment, inputStr string) (fdeployment.ChangesetOutput, error)
+	configuration(input string) (any, error)
 }
 
 type ChangeSet internalChangeSet
@@ -333,6 +334,17 @@ func (ccs ChangeSetImpl[C]) Configurations() (Configurations, error) {
 		ConfigResolver:      ccs.ConfigResolver,
 		InputType:           inputType,
 	}, nil
+}
+
+func (ccs ChangeSetImpl[C]) configuration(input string) (any, error) {
+	if input != "" && ccs.configProviderWithInput != nil {
+		return ccs.configProviderWithInput(input)
+	}
+	if ccs.configProvider != nil {
+		return ccs.configProvider()
+	}
+
+	return nil, errors.New("no configuration provider found")
 }
 
 // WithPreHooks appends pre-hooks to this changeset. Multiple calls are additive.
--- a/engine/cld/changeset/common.go
+++ b/engine/cld/changeset/common.go
@@ -296,13 +296,13 @@ func (ccs ChangeSetImpl[C]) Apply(env fdeployment.Environment) (fdeployment.Chan
 }
 
 func (ccs ChangeSetImpl[C]) applyWithInput(env fdeployment.Environment, input any) (fdeployment.ChangesetOutput, error) {
+	if reflect.ValueOf(input).IsZero() {
+		return ccs.Apply(env)
+	}
+
 	cInput, ok := input.(C)
 	if !ok {
 		return fdeployment.ChangesetOutput{}, fmt.Errorf("invalid input type: expected %T but got %T", *new(C), input)
-	}
-
-	if reflect.ValueOf(input).IsZero() {
-		return ccs.Apply(env)
 	}
 
 	if err := ccs.changeset.operation.VerifyPreconditions(env, cInput); err != nil {
--- a/engine/cld/changeset/common.go
+++ b/engine/cld/changeset/common.go
@@ -33,8 +33,8 @@ type internalChangeSet interface {
 	noop() // unexported function to prevent arbitrary structs from implementing ChangeSet.
 	Apply(env fdeployment.Environment) (fdeployment.ChangesetOutput, error)
 	Configurations() (Configurations, error)
-	applyWithInput(env fdeployment.Environment, inputStr string) (fdeployment.ChangesetOutput, error)
-	configuration(input string) (any, error)
+	applyWithInput(env fdeployment.Environment, input any) (fdeployment.ChangesetOutput, error)
+	resolvedInput(input string) (any, error)
 }
 
 type ChangeSet internalChangeSet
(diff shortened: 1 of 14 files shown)


engine/cld/commands/pipeline/run.go
```diff
@@ -221,13 +229,15 @@ func saveChangesetProposalMetadata(changesetName string, out fdeployment.Changes
...
+			Config any             `json:"config"`
 		}{{
-			ID:    id,
-			Name:  changesetName,
-			Input: json.RawMessage(changesetInputJSON),
+			ID:     id,
+			Name:   changesetName,
+			Input:  json.RawMessage(changesetInputJSON),
+			Config: changesetConfig,
 		}}
```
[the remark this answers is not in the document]
gustavogama-cll: > json.Number marshals as a JSON string under

that's not true

engine/cld/changeset/common.go
```diff
@@ -32,8 +32,9 @@ type Configurations struct {
 type internalChangeSet interface {
 	noop() // unexported function to prevent arbitrary structs from implementing ChangeSet.
 	Apply(env fdeployment.Environment) (fdeployment.ChangesetOutput, error)
-	applyWithInput(env fdeployment.Environment, inputStr string) (fdeployment.ChangesetOutput, error)
 	Configurations() (Configurations, error)
+	applyWithInput(env fdeployment.Environment, inputStr string) (fdeployment.ChangesetOutput, error)
+	configuration(input string) (any, error)
```
gustavogama-cll: I don't like the name of this method (as we already have `Configurations` with very different semantic. But I can't think of anything better.

gustavogama-cll: also: `input` is an optional string, which is passed to `configProviderWithInput` if one if provided. It is ok to have input be optional by handling `""` as no input, or should we use functional options (i.e., `WithInput()`?

engine/cld/changeset/registry.go
```diff
@@ -339,6 +346,17 @@ func (r *ChangesetsRegistry) GetChangesetOptions(key string) (ChangesetConfig, e
 	return entry.options, nil
 }
 
+// GetConfiguration retrieves the configuration for a changeset.
+// \"input\" is the optional input string passed to \"registry.Apply()\".
+func (r *ChangesetsRegistry) GetConfiguration(key string, input string) (any, error) {
```
gustavogama-cll: same points raised about `internalChangeset.configuration` apply to this one

engine/cld/changeset/common.go
graham-chainlink: In hindsight, i should have introduce the WIthInput functional option to `Apply` , instead of introducing `applyWithInput`

graham-chainlink: Hmm the 2 configuration name is definitely confusing, what about calling it finalInput or resolvedInput or processedInput configuredInput , buildInput?

engine/cld/changeset/registry.go
```diff
@@ -217,6 +217,11 @@ func (r *ChangesetsRegistry) Apply(
 		return fdeployment.ChangesetOutput{}, err
 	}
 
+	changesetConfig, err := applySnapshot.registryEntry.changeset.configuration(cfg.inputStr)
```
graham-chainlink: hmm this means the input is actually resolved twice, once in `configuration` , then another in `applyWithInput` , just slight in efficiency, i wonder if we can build the input and pass that into `applyWithInput` instead, then `applyWithInput` dont have to resolve the final input it self? So we resolve the input earlier before the prehook instead of currently in the apply phase.

Is it worth doing i wonder? What do you think?

engine/cld/changeset/common.go
gustavogama-cll: sure, I'll pick one of these. I avoided "input" as a name because in the changeset type definitions they are always referred to as "config".

engine/cld/changeset/registry.go
gustavogama-cll: sure, let me give it a try

engine/cld/changeset/common.go
gustavogama-cll: renamed to `resolvedInput` in 6f38d6a434f6788a12a50de123dbfcdbc31fc4b1

engine/cld/changeset/registry.go
gustavogama-cll: fixed in 6f38d6a434f6788a12a50de123dbfcdbc31fc4b1

engine/cld/changeset/common.go
```diff
@@ -294,23 +295,21 @@ func (ccs ChangeSetImpl[C]) Apply(env fdeployment.Environment) (fdeployment.Chan
...
-	if inputStr == "" {
-		return ccs.Apply(env)
+func (ccs ChangeSetImpl[C]) applyWithInput(env fdeployment.Environment, input any) (fdeployment.ChangesetOutput, error) {
+	cInput, ok := input.(C)
+	if !ok {
+		return fdeployment.ChangesetOutput{}, fmt.Errorf("invalid input type: expected %T but got %T", *new(C), input)
 	}
-	if ccs.configProviderWithInput == nil {
+
+	if reflect.ValueOf(input).IsZero() {
```
graham-chainlink: nitpick: i wonder if we should move this check to before line 299 as it doesnt need to work with `cInput`?

graham-chainlink (dismissed):
LGTM

engine/cld/changeset/common.go
gustavogama-cll: fixed in 5d81cad8126c1462fa1fc8bd6647b5e862aff54f

engine/cld/changeset/common.go
```diff
@@ -294,23 +295,21 @@ func (ccs ChangeSetImpl[C]) Apply(env fdeployment.Environment) (fdeployment.Chan
...
 		return ccs.Apply(env)
 	}
 
-	c, err := ccs.configProviderWithInput(inputStr)
-	if err != nil {
-		return fdeployment.ChangesetOutput{}, err
+	cInput, ok := input.(C)
+	if !ok {
+		return fdeployment.ChangesetOutput{}, fmt.Errorf("invalid input type: expected %T but got %T", *new(C), input)
 	}
```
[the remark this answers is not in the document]
graham-chainlink: Interesting I need to look into this more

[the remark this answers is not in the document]
gustavogama-cll: Copilot's right. Let me add a `input == nil` to the `if`.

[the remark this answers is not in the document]
gustavogama-cll: fixed in 51eb98732993aa237983acd7604f20a6c5c44e06

engine/cld/changeset/common.go
```diff
@@ -294,23 +295,21 @@ func (ccs ChangeSetImpl[C]) Apply(env fdeployment.Environment) (fdeployment.Chan
 	return ccs.changeset.operation.Apply(env, c)
 }
 
-func (ccs ChangeSetImpl[C]) applyWithInput(env fdeployment.Environment, inputStr string) (fdeployment.ChangesetOutput, error) {
-	if inputStr == "" {
-		return ccs.Apply(env)
-	}
-	if ccs.configProviderWithInput == nil {
+func (ccs ChangeSetImpl[C]) applyWithInput(env fdeployment.Environment, input any) (fdeployment.ChangesetOutput, error) {
+	if input == nil || reflect.ValueOf(input).IsZero() {
```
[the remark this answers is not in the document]
gustavogama-cll: @graham-chainlink , reviewing this yet again... I don't think it's possible to do what you suggested without changing `applyWithInput` a bit more. The information about whether `cfg.inputStr` was set or not is lost when we call `resolvedInput(cfg.inputStr)`. So the decision to call `Apply` or `applyWithInput` must be done by the caller (i.e., the code that has access to cfg.inputStr).

engine/cld/changeset/registry.go
```diff
@@ -339,6 +351,17 @@ func (r *ChangesetsRegistry) GetChangesetOptions(key string) (ChangesetConfig, e
...
 }
 
+// GetResolvedInput retrieves the configuration for a changeset.
+// \"input\" is the optional input string passed to \"registry.Apply()\".
+func (r *ChangesetsRegistry) GetResolvedInput(key string, input string) (any, error) {
+	entry, ok := r.entries[key]
+	if !ok {
+		return nil, fmt.Errorf("changeset '%s' not found", key)
+	}
+
```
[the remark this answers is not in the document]
gustavogama-cll: fix in 75a12bb3c8d63e45dfea662861821c3c9867589a

engine/cld/changeset/common.go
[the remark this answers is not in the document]
graham-chainlink: yeah, okay at least we tried

engine/cld/changeset/common.go
```diff
@@ -294,23 +295,17 @@ func (ccs ChangeSetImpl[C]) Apply(env fdeployment.Environment) (fdeployment.Chan
...
-func (ccs ChangeSetImpl[C]) applyWithInput(env fdeployment.Environment, inputStr string) (fdeployment.ChangesetOutput, error) {
-	if inputStr == "" {
-		return ccs.Apply(env)
-	}
-	if ccs.configProviderWithInput == nil {
-		return ccs.Apply(env)
+func (ccs ChangeSetImpl[C]) applyWithInput(env fdeployment.Environment, input any) (fdeployment.ChangesetOutput, error) {
+	cInput, ok := input.(C)
+	if !ok {
+		return fdeployment.ChangesetOutput{}, fmt.Errorf("invalid input type: expected %T but got %T", *new(C), input)
```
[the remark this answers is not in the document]
gustavogama-cll: What the heck, Copilot... there's no more "reflect.ValueOf" in the new code!

graham-chainlink (dismissed):

engine/cld/changeset/common.go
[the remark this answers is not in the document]
graham-chainlink: Yeap it gets confused sometimes , usualy happen when i force push or push many times in short period

graham-chainlink (approved):

engine/cld/changeset/common.go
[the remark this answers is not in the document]
gustavogama-cll: @graham-chainlink , one more try...
Tests still pass with this latest version, but `Apply` is now useless (it's only used in tests).

graham-chainlink (approved):

PR #925, opened 2026-04-13
[CLD-1983]: feat(operations): add WithForceExecute option and latest cache reuse

graham-chainlink:
## What changed
- Added `WithForceExecute` execute option to force operation execution and bypass previous successful report short-circuiting.
- Updated cached report lookup to prefer the most recent matching successful report.

## Why
- We have use case from our users where in the same changeset , they want to execute the same operations but it can return different data due to it being update in between calls. With the operations report caching, they will never be able to fetch the latest data, introducing an option to bypass the cache will support this use case.

Context: https://chainlink-core.slack.com/archives/C043E82DUM8/p1775799230707799

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-1983

### DIFF
--- a/.changeset/gentle-moles-rule.md
+++ b/.changeset/gentle-moles-rule.md
@@ -1,0 +1,12 @@
+---
+"chainlink-deployments-framework": minor
+---
+
+feat(operations-api): introduce `WithForceExecute` and new `ExecuteOperationN` options
+
+### Migration: `ExecuteOperationN`
+
+**Signature**
+
+- Before: `ExecuteOperationN(..., opts ...ExecuteOption[IN, DEP])`
+- After: `ExecuteOperationN(..., opts ...ExecuteOperationNOption[IN, DEP])`
--- a/operations/execute.go
+++ b/operations/execute.go
@@ -13,9 +13,19 @@ var ErrNotSerializable = errors.New("data cannot be safely written to disk witho
 // ExecuteConfig is the configuration for the ExecuteOperation function.
 type ExecuteConfig[IN, DEP any] struct {
 	retryConfig RetryConfig[IN, DEP]
+	// forceExecute controls whether execution should skip execution when previous successful report is found (set by WithForceExecute).
+	forceExecute bool
 }
 
 type ExecuteOption[IN, DEP any] func(*ExecuteConfig[IN, DEP])
+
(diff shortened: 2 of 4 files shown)


operations/execute.go
```diff
@@ -355,7 +365,8 @@ func loadPreviousSuccessfulReport[IN, OUT any](
 		return Report[IN, OUT]{}, false
 	}
 
-	for _, report := range prevReports {
+	for i := len(prevReports) - 1; i >= 0; i-- {
```
graham-chainlink, line 400: now that there can be multiple reports of the same ID, we want to return the most recent ones instead of the oldest one

.changeset/gentle-moles-rule.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+feat(operations-api): introduce WithAlwaysExecute and new ExecuteOperationN options
+
+### Migration: `ExecuteOperationN`
+
+**Signature**
+
+- Before: `ExecuteOperationN(..., opts ...ExecuteOption[IN, DEP])`
+- After: `ExecuteOperationN(..., opts ...ExecuteOperationNOption[IN, DEP])`
```
graham-chainlink, line 12: This is a breaking change, but i checked company wide usage for `ExecuteOperationN` and didn see any.

operations/execute.go
```diff
@@ -13,10 +13,20 @@ var ErrNotSerializable = errors.New("data cannot be safely written to disk witho
...
 type ExecuteConfig[IN, DEP any] struct {
 	retryConfig RetryConfig[IN, DEP]
+	// alwaysExecute controls whether execution should skip report cache reuse (set by WithAlwaysExecute).
+	alwaysExecute bool
 }
 
 type ExecuteOption[IN, DEP any] func(*ExecuteConfig[IN, DEP])
 
+// ExecuteOperationNConfig holds options for ExecuteOperationN.
+type ExecuteOperationNConfig[IN, DEP any] struct {
```
graham-chainlink, line 23: I have to introduce new functional option type for `ExecuteOperationN` as i dont want `ExecuteOperationN` to support always run option as it has its own quirks. Previously it was sharing the same option type as `ExecuteOperation`

ajaskolski (approved):

jkongie (approved):

PR #927, opened 2026-04-14
ci(operations-gen): add script to perform tagging and release

graham-chainlink:
Add `tools/operations-gen/release-tag.sh` for automating tagging and release as a temporary measure to start with

I expect once we introduce [release-it](https://github.com/release-it/release-it) in this repo, we can further automate this in CI.

```
go install github.com/smartcontractkit/chainlink-deployments-framework/tools/operations-gen@v0.0.1
```

Screenshot 2026-04-14 at 4 44 03 pm

### DIFF
--- a/tools/operations-gen/release-tag.sh
+++ b/tools/operations-gen/release-tag.sh
@@ -1,0 +1,163 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+usage() {
+  cat <<'EOF'
+Create, push, and optionally publish a GitHub release for operations-gen.
+
+Usage:
+  ./tools/operations-gen/release-tag.sh v0.1.0 [--push] [--release] [--allow-dirty]
+
+Examples:
+  ./tools/operations-gen/release-tag.sh v0.1.0
+  ./tools/operations-gen/release-tag.sh v0.1.0 --push
+  ./tools/operations-gen/release-tag.sh v0.1.0 --release
+
+Notes:
+  - Tag format is always: tools/operations-gen/<version>
+  - --release implies --push and uses GitHub CLI (gh)
+  - Release notes are scoped to commits touching tools/operations-gen
+  - Use this tag for go install:
+    go install github.com/smartcontractkit/chainlink-deployments-framework/tools/operations-gen@v0.1.0
+EOF
+}
+
+if [[ "${1:-}" == "-h" || "${1:-}" == "--help" || $# -lt 1 ]]; then
+  usage
+  exit 0
+fi
+
+version="$1"
+shift
+
+push_tag="false"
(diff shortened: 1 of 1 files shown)


tools/operations-gen/release-tag.sh
DimitriosNaikopoulos, line 1: Is there a reason why we do not use changeset releases for this tool since the repo already supports that 🤔 Also are we only plan to manually release new versions?

tools/operations-gen/release-tag.sh
```diff
@@ -0,0 +1,163 @@
...
+      exit 1
+      ;;
+  esac
+done
+
+if [[ "$create_release" == "true" ]]; then
+  push_tag="true"
+fi
+
+if [[ ! "$version" =~ ^v[0-9]+\.[0-9]+\.[0-9]+([\-+][0-9A-Za-z\.-]+)?$ ]]; then
```
DimitriosNaikopoulos, line 54: Since this is manual we will need to keep track of version bumps and there is a risk two people releasing the same version for a different feature that they are working on. Although if we do not expect a lot of parallel work in the tool then this should not be an issue only a bit more difficult to release since you would have to know what was the latest version to manually bump it to the next.

DimitriosNaikopoulos (commented):
Is there a reason why we are not introducing changesets to handle releases as we do with the framework until we start using release please? 

Do we expect to make significant use of the script or we could just create manual releases as a stop gap? Do we need a changelog file?

tools/operations-gen/release-tag.sh
graham-chainlink: Maybe there is better way now, in Chainlink-protos we had to create own script to support monorepo - https://github.com/smartcontractkit/chainlink-protos/blob/main/scripts/tag-and-push.sh which i didn quite like, not to mention new package.json inside operations-gen folder.

So i was thinking with Operations-gen, we use script release in the mean time until we move to release-it, i expect very little activity for operations-gen, if it picks up, then we can just migrate to release-it for CLDF.

DimitriosNaikopoulos: Actually the [changeset action](https://github.com/smartcontractkit/.github/blob/main/actions/cicd-changesets/action.yml) supports mono repo and multi package releases. You can have the root to keep using the v<semver> format and any other package using a `tool-name/v<semver>` format. 

I just don't like custom release scripts either because they make releases unpredictable from repo-to-repo and harder to maintain. 

If this is a stop gap till we move the whole repo to release it (does release it support multi package releases from the same repo) then we can keep the script. If though we are keeping changeset releases for cldf then we should migrate this to a changeset release since there is support for it.

PR #930, opened 2026-04-15
[CLD-1980]: ci: setup release pipeline for operations-gen

graham-chainlink:
Setup changeset config and go releaser config

Similar to [chainlink-protos here](https://github.com/smartcontractkit/chainlink-protos/blob/main/package.json#L5), i have to override the `ci:changeset:publish` command with custom script. This is because changeset by default does not understand the tagging semantic for go.

Per the [go mod docs](https://go.dev/ref/mod#vcs-version), 

> If a module is defined in a subdirectory within the repository, that is, the [module subdirectory](https://go.dev/ref/mod#glos-module-subdirectory) portion of the module path is not empty, then each tag name must be prefixed with the module subdirectory, followed by a slash. For example, the module golang.org/x/tools/gopls is defined in the gopls subdirectory of the repository with root path golang.org/x/tools. The version v0.4.0 of that module must have the tag named gopls/v0.4.0 in that repository.

the tag has to be the format `<path-to-gomod>/v0.4.0`. eg instead of `operations-gen@0.4.0` , we want `tools/operations-gen/v0.4.0`. This means we have to perform tagging ourselves

JIRA:  https://smartcontract-it.atlassian.net/browse/CLD-1980

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "A deployment framework for chainlink-deployments ",
   "private": true,
   "scripts": {
-    "ci:changeset:publish": "pnpm changeset publish",
+    "ci:changeset:publish": "node ./scripts/release/tag-releases.mjs",
     "ci:changeset:version": "pnpm changeset version && pnpm version --patch"
   },
   "author": "@smartcontractkit",
--- a/tools/operations-gen/package.json
+++ b/tools/operations-gen/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "operations-gen",
+  "version": "0.0.1",
+  "private": true,
+  "description": "Changesets metadata package for tools/operations-gen Go module"
+}
--- a/.github/workflows/push-tag-release-operations-gen.yml
+++ b/.github/workflows/push-tag-release-operations-gen.yml
@@ -1,0 +1,27 @@
+name: push-tag-release-operations-gen
+
+on:
+  push:
+    tags:
+      - tools/operations-gen/v*
+
+jobs:
+  cicd-publish-release-operations-gen:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: write
+      actions: read
+    steps:
+      - name: cicd-publish-release-operations-gen
+        uses: smartcontractkit/.github/actions/cicd-build-publish-artifacts-go@b4737861584f88fa9569d6978f70fedf8b1ae67c # cicd-build-publish-artifacts-go@0.4.0
+        with:
+          # general inputs
+          app-name: operations-gen
+          publish: "false" # do not publish docker image to ECR
+          update-git-tag: "false"
+          # goreleaser inputs
+          goreleaser-args: "--config tools/operations-gen/.goreleaser.yml"
+          goreleaser-version: '~> v2'
+          goreleaser-dist: goreleaser-pro
+          goreleaser-key: ${{ secrets.GORELEASER_KEY }}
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "A deployment framework for chainlink-deployments ",
   "private": true,
   "scripts": {
-    "ci:changeset:publish": "pnpm changeset publish",
+    "ci:changeset:publish": "node ./scripts/release/tag-releases.mjs",
```
graham-chainlink, line 7: Refer to PR description why we need to do this

tools/operations-gen/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "operations-gen",
```
DimitriosNaikopoulos, line 2: Is it possible to have the name as `tools/operations-gen` ? I am thinking if that works then we do not need the custom script in order to ensure that the release tag is valid for Go packages 🤔 Eg this could generate a `tools/operations-gen/vx.y.z` tag

graham-chainlink: I tried it , it will generate the right name eg `tools/operations-gen/0.4.0` but the version is missing `v` , and we want `tools/operations-gen/v0.4.0`, also when in this mono repo mode, changeset will also publish the root as `chainlink-deployments-framework@0.94.1` instead of just `v0.94.1`
However when it is in single repo mode, it generates the `v0.94.1` with v correctly, weird.

I jsut think that changeset doesnt work well with golang for `go get`

DimitriosNaikopoulos: I might be wrong but if you set the [separator](https://github.com/smartcontractkit/cld-actions/blob/44eaa36a6f260d974a58838481e7285fd678f3fe/.github/workflows/push-main.yml#L42) as /v it should have the correct format that is what we do in cld-action to get [releases](https://github.com/smartcontractkit/cld-actions/releases) with the `v` format

DimitriosNaikopoulos (approved):
I left a suggestion that you might have already tried therefore I will approve this so you are not blocked. In general it would be better if we could avoid the custom logic around tags.

tools/operations-gen/package.json
graham-chainlink: Oh nice, maybe that can fix the version format, but we still have the issue with the root go mod that changeset will try to publish `chainlink-deployments-framework@v0.94.1` instead of just `v0.94.1` once we configure the mono repo mode with workspace

graham-chainlink: You can also see that in cld-actions, the package on the root is release as `cld-actions/vX.X.X` instead of `v.X.X.X` that we want in cldf

DimitriosNaikopoulos: This is actually because we do not set this in cld-action https://github.com/smartcontractkit/.github/blob/4475ad7c475bacb1aefeccb132645149201ababd/actions/cicd-changesets/action.yml#L73-L78

So the release is using the full package name `cld-action/vx.y.z`

graham-chainlink: For historical reference, pasting my findings here on why we can use cicd-changesets with current setup
------
i can get almost there, tags are correct but [the workflow will failed](https://github.com/smartcontractkit/chainlink-deployments-framework/actions/runs/24494328290/job/71585782216#step:2:1246) because both changesets-tag-separator and changesets-root-version-package-path  do not seem to work together, there is a piece of logic in signed commit action that is not handling this scenario correctly

we would need to update signed commit action, the cicd changeset action and finally we can use them in our repo. But i will go with my script for now and may put it a bug fix for the .github actions when i get some time.

Screenshot 2026-04-16 at 4 36 33 pm

PR #937, opened 2026-04-16
update ctf framework and sui version

FelixFan1992:


chain/sui/provider/ctf_provider.go
```diff
@@ -36,7 +36,7 @@ type CTFChainProviderConfig struct {
 	Once *sync.Once
 
 	// Optional: A specification of the image to use for the CTF container.
-	// Default: mysten/sui-tools:devnet
+	// Default: mysten/sui-tools:ci-arm64 on darwin/arm64; mysten/sui-tools:devnet-v1.69.2 elsewhere.
```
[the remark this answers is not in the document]
gustavogama-cll: I think this is worth addressing

chain/sui/provider/ctf_provider.go
```diff
@@ -184,29 +184,28 @@ func (p *CTFChainProvider) startContainer(
...
 		if p.config.Platform != nil {
-			platform = *p.config.Platform
-		} else {
-			if runtime.GOARCH == "arm64" {
-				platform = "linux/arm64"
-			}
+			imagePlatform = p.config.Platform
+		} else if runtime.GOARCH == "arm64" {
+			arm := "linux/arm64"
+			imagePlatform = &arm
```
gustavogama-cll: (nit)
```suggestion
			imagePlatform = pointer.To("linux/arm64")
```

(`pointer` is from "github.com/smartcontractkit/chainlink-deployments-framework/internal/pointer"; after we move to go 1.26 we can use `new`)

chain/sui/provider/ctf_sui_blockchain.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+
+	"github.com/smartcontractkit/chainlink-testing-framework/framework"
+	"github.com/smartcontractkit/chainlink-testing-framework/framework/components/blockchain"
+	"github.com/smartcontractkit/chainlink-testing-framework/framework/pods"
+)
+
+// defaultSuiImage is the mysten/sui-tools image used when CTFChainProviderConfig.Image is unset
+// (non–darwin/arm64 hosts). Keep aligned with the Sui release you support in tests.
```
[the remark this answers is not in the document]
gustavogama-cll: same comment here; probably worth addressing Copilot's point

chain/sui/provider/ctf_sui_blockchain.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	if in.Port == "" {
+		in.Port = blockchain.DefaultSuiNodePort
+	}
+	if in.FaucetPort == "" {
+		in.FaucetPort = blockchain.DefaultFaucetPortNum
+	}
+}
+
+// newSuiCTFBlockchainNetwork mirrors CTF blockchain.newSui but uses generateKeyDataCTF so
```
gustavogama-cll, line 111: why can't we use/patch CTF instead? I don't think we should have this much container management inside CLDF unless we really must.

chain/sui/provider/ctf_provider.go
FelixFan1992: done

chain/sui/provider/ctf_sui_blockchain.go
FelixFan1992: will address this in CTF first

graham-chainlink: agree, if we bake this is CTF, then more people can benefit it too?

PR #938, opened 2026-04-16
Fix/update anvil startup

AnieeG:
Using the official image for anvil
Adding `--no-storage-caching` option to ensure state is always fetched from onchain instead of operating on local cached state . The local cached state can lead to `missing trie node ... state ... not available` if the rpc used for forking does not return the full chain state


engine/cld/environment/anvil.go
```diff
@@ -203,7 +203,7 @@ func newAnvilChains(
 			}
 			network.Metadata = cfgnet.EVMMetadata{
 				AnvilConfig: &cfgnet.AnvilConfig{
-					Image: "f4hrenh9it/foundry:latest",
+					Image: "ghcr.io/foundry-rs/foundry:latest",
```
gustavogama-cll, line 206: @skudasov , do you see any issues with this?

gustavogama-cll (dismissed):

gustavogama-cll (approved):

engine/cld/environment/anvil.go
skudasov: We are using `stable` in CTF by default https://github.com/smartcontractkit/chainlink-testing-framework/blob/25511053de570f98ceab034dfbf78e493dd56962/framework/components/blockchain/anvil.go#L30, but I think if it works for you, it works :)

PR #942, opened 2026-04-18
feat(mcms/analyzer): add Solana native programs to decoder registry

gustavogama-cll:
This pull request enhances the Solana analyzer by adding support for native Solana programs and improving instruction decoding capabilities. It introduces a registry for Solana native programs, ensures their decoders are registered by default, and adds comprehensive support for Anchor IDL instruction decoding.

It also adds Go bindings for some native programs that are not available by default in the gagliardetto/solana-go library to the `experimental/analyzer/solana/programs` package. 

---
OPT-486


experimental/analyzer/solana/programs/bpfloader/instructions.go
```diff
@@ -0,0 +1,153 @@
...
+func registryDecodeInstruction(accounts []*ag_solanago.AccountMeta, data []byte) (any, error) {
+	inst, err := DecodeInstruction(accounts, data)
+	if err != nil {
+		return nil, err
+	}
+
+	return inst, nil
+}
+
+func DecodeInstruction(accounts []*ag_solanago.AccountMeta, data []byte) (*Instruction, error) {
```
ecPablo: should we add some unit tests to these ? not sure how easy it is to generate meaningful test data for this one but flagging in case we can

experimental/analyzer/solana/programs/idl/instructions.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// IsInstruction reports whether data starts with the IDL instruction discriminator.
+func IsInstruction(data []byte) bool {
+	return len(data) >= 9 && bytes.Equal(data[:8], Discriminator)
+}
+
+// DecodeInstruction decodes a Solana IDL instruction from the given accounts and data.
+// The data must start with the 8-byte IDL discriminator followed by a 1-byte instruction type.
+func DecodeInstruction(accounts []*ag_solanago.AccountMeta, data []byte) (*Instruction, error) {
```
ecPablo, line 88: Same for this one , maybe we can add some unit tests

experimental/analyzer/solana/programs/loaderv3/accounts.go
```diff
@@ -0,0 +1,3 @@
+// Code generated by https://github.com/gagliardetto/anchor-go. DO NOT EDIT.
+
+package loaderv3
```
ecPablo, line 3: For all of these, I'm wondering if we should add some `Taskfile` on this repo to re generate them in case someone needs to or there is some program version update? Just so other devs have context on exactly what where the source programs that generated these bindings

experimental/analyzer/solana/programs/bpfloader/instructions.go
gustavogama-cll: The bindings are generated. However, this one (bpfloader) slipped by and shouldn't have been added (not sure why Copilot didn't flag it).

experimental/analyzer/solana/programs/loaderv3/accounts.go
gustavogama-cll: we don't have the source code for these easily, so it might get more complicated than it's warranted. But I'll try to add some instructions and references to the source of the IDL file.

experimental/analyzer/solana/programs/idl/instructions.go
gustavogama-cll: ok

experimental/analyzer/registry_solana.go
```diff
@@ -84,6 +124,15 @@ func NewEnvironmentSolanaRegistry(env deployment.Environment, decoderMappings ma
 		addressesByChain[address.ChainSelector] = chainAddresses
 	}
 
+	// add native programs and mappings
+	for chainSelector, addresses := range addressesByChain {
```
ChrisAmora, line 130: Is it an invariant that every proposal target chain is present in the AddressBook/DataStore with at least one address entry?

experimental/analyzer/solana/programs/idl/instructions.go
gustavogama-cll: done 6a0cf665bc4e6d0b21095d0cd80a0d35b867302c

experimental/analyzer/registry_solana.go
gustavogama-cll: yes, you need at least the timelock and mcm addresses in the datastore/address book to generate a proposal

experimental/analyzer/registry_solana.go
```diff
@@ -84,6 +124,15 @@ func NewEnvironmentSolanaRegistry(env deployment.Environment, decoderMappings ma
...
+		for pk, contractTypeAndVersion := range nativePrograms {
+			addresses[pk.String()] = contractTypeAndVersion
+		}
+		addressesByChain[chainSelector] = addresses
+	}
+	maps.Insert(decoderMappings, maps.All(nativeDecoderMappings))
```
graham-chainlink: `decoderMappings` is passed into this method as parameter, and here we are mutating it, should we clone decoderMappings and then add the new mappings to it?

gustavogama-cll: I think it's unlikely to make a difference but I pushed a patch cloning the original mapping [here](https://github.com/smartcontractkit/chainlink-deployments-framework/pull/942/changes/efdeb868ec2e6b144385d4c8dd59f9281bb8fd0c)

graham-chainlink (approved):

PR #948, opened 2026-04-22
chore: port link token types

ecPablo:
This pull request introduces new contract type definitions for LINK tokens to the `engine/cld/link` package, aligning with the types used in the `chainlink-deployments-framework`. This update helps standardize LINK token references across deployments.

**Contract type definitions:**

* Added `LinkToken` and `StaticLinkToken` constants of type `cldf.ContractType` in `contract_types.go` to represent the standard burn/mint LINK token and the legacy static LINK token, respectively.

**Changelog:**

* Added a changeset entry documenting the minor version bump and the porting of LINK token types.


engine/cld/contracts/link/contract_types.go
graham-chainlink, line 1: almost feel like there should be another folder between cld and link , but i am not sure what to call it, having link under engine/cld seems a bit awkward to me , what do you think?

ChrisAmora (dismissed):

engine/cld/contracts/link/contract_types.go
ecPablo: Yeah maybe we can move have one call contract_types. We already have some others under the proposalutils pkg https://github.com/smartcontractkit/chainlink-deployments-framework/blob/main/engine/cld/mcms/proposalutils/contract_types.go maybe we can move those too and have them all together. Or something more generic like `constants`

graham-chainlink: something like eg `contract/types.go` instead of constants?

ecPablo: yeah that could work, we have contract/mcms/types.go, contract/link/types.go?

ecPablo: Added this: https://github.com/smartcontractkit/chainlink-deployments-framework/pull/948/commits/9d16bb2ef26b4479235efa12ec7b7a7189e3fec3 just noting that this is a breaking  change as all the core imports of these types will have to be changed when we bump cldf on core since we already replaced a bunch of usages here: https://github.com/smartcontractkit/chainlink/pull/22019

ChrisAmora (approved):

PR #954, opened 2026-04-23
CLD-2058: Add in memory and file catalog deletes

DimitriosNaikopoulos:
# CLD-2058: Add in memory and file catalog deletes

- Adds Deletions keys for `AddressRef`, `ChainMetadata` and `ContractMetadata`
- Remove deleted records and save keys for later removing them from catalog
- Add a Method that can convert the deletion key string to a private `keyRef` to avoid exposing the struct  and allow for key deletions.
- During merge, ensure that after the two datastores are merged, we call delete for any key that is marked for deletion. 
- If a key does not exist and we try to delete it return a not found error informing the user about a potential bug in their changeset or wrong input. 

## Example

Directly delete from in-memory datastore in a changeset (only changeset related contract deployments available). This instanyl

```
	dataStore.Addresses().Delete(ds.AddressRefKey{
		ChainSelector: chainSelector,
		Type:          "LinkToken",
		Version:       semver.MustParse("1.0.0"),
		Qualifier:     "LinkTokenContractV1_" + addr.String(),
	})
```

Staging a file for deletion for a file that exists in the inflated datastore (remote datastore either backed by file or catalog).

```
 # Fetch all addresses
	addresses, err := e.DataStore.Addresses().Fetch()
	if err != nil {
		return cldf.ChangesetOutput{},
			fmt.Errorf("failed to fetch addresses: %w", err)
	}

# Find specific address 
	for _, address := range addresses {
		if address.Address == "0x123" {
# Stage the deletion for a later step (during merge)
			dataStore.Addresses().RemoteDelete(address.Key())
		}
	}

### DIFF
--- a/datastore/address_ref_key_test.go
+++ b/datastore/address_ref_key_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestAddressRefKey_Equals(t *testing.T) {
@@ -75,4 +76,55 @@ func TestAddressRefKey_String(t *testing.T) {
 	key := NewAddressRefKey(42, ContractType("MyType"), semver.MustParse("1.2.3"), "qual")
 	expected := "42_MyType_1.2.3_qual"
 	assert.Equal(t, expected, key.String())
+}
+
+func TestNewAddressRefKeyFromString(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		give    string
+		want    AddressRefKey
+		wantErr string
+	}{
+		{
+			name: "success: valid string",
+			give: "42_MyContract_1.2.3_primary",
+			want: NewAddressRefKey(42, ContractType("MyContract"), semver.MustParse("1.2.3"), "primary"),
+		},
+		{
+			name:    "failure: too few parts",
+			give:    "42_MyContract_1.2.3",
+			wantErr: "invalid address ref key",
+		},
+		{
+			name:    "failure: too many parts",
+			give:    "42_MyContract_1.2.3_primary_extra",
+			wantErr: "invalid address ref key",
+		},
+		{
+			name:    "failure: invalid chain selector",
+			give:    "notanumber_MyContract_1.2.3_primary",
+			wantErr: "failed to parse chain selector",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			got, err := NewAddressRefKeyFromString(tt.give)
+
+			if tt.wantErr != "" {
+				require.ErrorContains(t, err, tt.wantErr)
+				return
+			}
+
+			require.NoError(t, err)
+			assert.Equal(t, tt.want.ChainSelector(), got.ChainSelector())
+			assert.Equal(t, tt.want.Type(), got.Type())
+			assert.True(t, tt.want.Version().Equal(got.Version()))
+			assert.Equal(t, tt.want.Qualifier(), got.Qualifier())
+		})
+	}
 }
--- a/datastore/chain_metadata_key.go
+++ b/datastore/chain_metadata_key.go
@@ -37,6 +37,19 @@ func (c chainMetadataKey) String() string {
 	return strconv.FormatUint(c.chainSelector, 10)
 }
 
+// NewChainMetadataKeyFromString creates a new ChainMetadataKey instance from a string representation.
+func NewChainMetadataKeyFromString(key string) (ChainMetadataKey, error) {
+	if len(key) == 0 {
+		return nil, fmt.Errorf("invalid chain metadata key: %s", key)
+	}
+	chainSelector, err := strconv.ParseUint(key, 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse chain selector: %w", err)
+	}
+
+	return NewChainMetadataKey(chainSelector), nil
+}
+
 // NewChainMetadataKey creates a new ChainMetadataKey instance.
 func NewChainMetadataKey(chainSelector uint64) ChainMetadataKey {
 	return chainMetadataKey{
--- a/datastore/contract_metadata_key.go
+++ b/datastore/contract_metadata_key.go
@@ -1,6 +1,10 @@
 package datastore
 
-import "fmt"
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
 
 // ContractMetadataKey is an interface that represents a key for ContractMetadata records.
 // It is used to uniquely identify a record in the ContractMetadataStore.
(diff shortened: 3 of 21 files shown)

### CODE REVIEW

File: datastore/memory_datastore.go [hunk 1]
```diff
@@ -55,39 +58,84 @@ func (s *MemoryDataStore) EnvMetadata() MutableEnvMetadataStore {
...
+		for _, dk := range src.DeletedRemoteKeys {
+			key, keyErr := NewContractMetadataKeyFromString(dk)
+			if keyErr != nil {
+				return fmt.Errorf("failed to parse contract metadata key: %w", keyErr)
+			}
+			if err = s.ContractMetadataStore.Delete(key); err != nil {
+				return err
+			}
+		}
+	}
```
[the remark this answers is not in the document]
DimitriosNaikopoulos: If the deletion fail at a later stage the changes will not be committed to the datastore

datastore/address_ref_key_test.go
```diff
@@ -76,3 +77,54 @@ func TestAddressRefKey_String(t *testing.T) {
...
+			name:    "failure: too many parts",
+			give:    "42_MyContract_1.2.3_primary_extra",
+			wantErr: "invalid address ref key",
+		},
+		{
+			name:    "failure: invalid chain selector",
+			give:    "notanumber_MyContract_1.2.3_primary",
+			wantErr: "failed to parse chain selector",
+		},
+	}
```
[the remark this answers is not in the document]
DimitriosNaikopoulos: AddressRef keys should automatically have a semver versioning based on the interface. If there is an error on what the user is providing, `semver.MustParse` will panic and inform the user about the problem

datastore/memory_datastore.go
```diff
@@ -55,39 +58,84 @@ func (s *MemoryDataStore) EnvMetadata() MutableEnvMetadataStore {
...
+		for _, dk := range src.DeletedRemoteKeys {
+			key, keyErr := NewAddressRefKeyFromString(dk)
+			if keyErr != nil {
+				return fmt.Errorf("failed to parse address ref key: %w", keyErr)
+			}
+			if err = s.AddressRefStore.Delete(key); err != nil {
+				return err
+			}
+		}
+	}
```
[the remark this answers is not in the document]
DimitriosNaikopoulos: Merge happens sequentially this repeats the same patern as upsert

datastore/memory_datastore.go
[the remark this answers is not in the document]
DimitriosNaikopoulos: Merge happens sequentially this repeats the same patern as upsert

datastore/memory_address_ref_store.go
ajaskolski, line 142: Yo left same comment in the doc.
RemoteDelete + Upsert leaving a stale entry in DeletedRemoteKeys
if this key is in DeletedRemoteKeys, it should be removed on successful Upsert. 
As if you then Upsert / Add that same key again without removing it from DeletedRemoteKeys, merge can still run Delete for that staged key after upserts and wipe the row that was just added back.

datastore/chain_metadata_key.go
```diff
@@ -37,6 +37,19 @@ func (c chainMetadataKey) String() string {
 	return strconv.FormatUint(c.chainSelector, 10)
 }
 
+// NewChainMetadataKeyFromString creates a new ChainMetadataKey instance from a string representation.
+func NewChainMetadataKeyFromString(key string) (ChainMetadataKey, error) {
+	if len(key) == 0 {
+		return nil, fmt.Errorf("invalid chain metadata key: %s", key)
+	}
+	chainSelector, err := strconv.ParseUint(key, 10, 64)
```
giogam, line 45: Does chainselectors lib offer conversion utils? If so we should consider using these

datastore/contract_metadata_key.go
```diff
@@ -41,6 +45,20 @@ func (c contractMetadataKey) String() string {
 	return fmt.Sprintf("%d_%s", c.chainSelector, c.address)
 }
 
+// NewContractMetadataKeyFromString creates a new ContractMetadataKey instance from a string representation.
+func NewContractMetadataKeyFromString(key string) (ContractMetadataKey, error) {
+	parts := strings.Split(key, "_")
+	if len(parts) != 2 {
+		return nil, fmt.Errorf("invalid contract metadata key: %s", key)
+	}
+	chainSelector, err := strconv.ParseUint(parts[0], 10, 64)
```
giogam, line 54: Same here, also whenever we parse a string into a chain selector, should we also validate that the chain selector is well-formed? It’s a public API, so once we ship it anyone can call it.

datastore/memory_address_ref_store.go
```diff
@@ -147,3 +149,21 @@ func (s *MemoryAddressRefStore) Delete(key AddressRefKey) error {
...
+func (s *MemoryAddressRefStore) RemoteDelete(key AddressRefKey) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	deletedKey := key.String()
+
+	if slices.Contains(s.DeletedRemoteKeys, deletedKey) {
+		return nil
+	}
+	s.DeletedRemoteKeys = append(s.DeletedRemoteKeys, deletedKey)
```
giogam:
```suggestion
if !slices.Contains(s.DeletedRemoteKeys, deletedKey) {
	s.DeletedRemoteKeys = append(s.DeletedRemoteKeys, deletedKey)
}
```

datastore/memory_chain_metadata_store.go
```diff
@@ -157,3 +159,21 @@ func (s *MemoryChainMetadataStore) Delete(key ChainMetadataKey) error {
...
+func (s *MemoryChainMetadataStore) RemoteDelete(key ChainMetadataKey) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	deletedKey := key.String()
+
+	if slices.Contains(s.DeletedRemoteKeys, deletedKey) {
+		return nil
+	}
+	s.DeletedRemoteKeys = append(s.DeletedRemoteKeys, deletedKey)
```
giogam: if !slices.Contains(s.DeletedRemoteKeys, deletedKey) {
	s.DeletedRemoteKeys = append(s.DeletedRemoteKeys, deletedKey)
}

datastore/memory_contract_metadata_store.go
```diff
@@ -157,3 +159,21 @@ func (s *MemoryContractMetadataStore) Delete(key ContractMetadataKey) error {
...
+func (s *MemoryContractMetadataStore) RemoteDelete(key ContractMetadataKey) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	deletedKey := key.String()
+
+	if slices.Contains(s.DeletedRemoteKeys, deletedKey) {
+		return nil
+	}
+	s.DeletedRemoteKeys = append(s.DeletedRemoteKeys, deletedKey)
```
giogam:
```suggestion
if !slices.Contains(s.DeletedRemoteKeys, deletedKey) {
	s.DeletedRemoteKeys = append(s.DeletedRemoteKeys, deletedKey)
}
```

DimitriosNaikopoulos: 🤦‍♂️ Yeah that makes more sense 😄 This logic is removed locally I start using a map[string]bool instead which is easier to handle deletes to address removing a deletion key if is added back in the datastore as @ajaskolski suggested

datastore/contract_metadata_key.go
DimitriosNaikopoulos: This is currently how we convert string values to chainselector values. We could check if the string is actually a valid chainselector but this breaks a lot of tests in our side were we use dummy chain selectors and I am not sure if it will also break tests client side if they are using dummy datastore values for testing.

datastore/chain_metadata_key.go
DimitriosNaikopoulos: same as below

datastore/memory_address_ref_store.go
DimitriosNaikopoulos: Fixed [ce058c0](https://github.com/smartcontractkit/chainlink-deployments-framework/pull/954/commits/ce058c0727597a5ad9789cf5cb26d319d18b7c87)

Took a couple of tries in order to ensure that we do not panic when we receive nil as the semver version for the address ref key. If we want to remove nil (*semver) from the address ref key interface this will be a breaking change (need to update this methods return type https://github.com/smartcontractkit/chainlink-deployments-framework/blob/CLD-2058-datastore-deletes/datastore/address_ref_key.go#L23)

ajaskolski (approved):

giogam (approved):

PR #959, opened 2026-04-24
feat: add DataStore and BlockChains to HookEnv type

gustavogama-cll:
This pull request expands the hook environment structures and related function calls to include access to `BlockChains` and `DataStore`. These additions ensure that both pre- and post-hooks, as well as proposal hooks, have access to these resources, enabling more comprehensive and integrated hook logic.

Because the BlockChains instance passed to hooks should not allow "writes", the PR also adds a `ReadOnly` method to the `Chain` interface and corresponding implementations to the existing chain families.

---
CLD-1965


chain/canton/canton_chain.go
```diff
@@ -150,3 +150,7 @@ func CreateAdminServiceClients(conn grpc.ClientConnInterface) AdminServiceClient
 		TrafficControl:           participantv30.NewTrafficControlServiceClient(conn),
 	}
 }
+
+func (c Chain) ReadOnly() (chaincommon.BlockChain, error) {
+	return c, nil
```
ecPablo, line 155: I saw in other chains we set this to a runtime generated key. Do we need to do the same here or maybe add a not implemented warning if we're still missing the correct way to do it for canton?

chain/internal/common/chain_metadata.go
```diff
@@ -9,6 +9,28 @@ import (
 	"github.com/smartcontractkit/chainlink-deployments-framework/chain/utils"
 )
 
+type BlockChain interface {
```
ecPablo, line 12: Is there a reason why we don't use the interface here? https://github.com/smartcontractkit/chainlink-deployments-framework/blob/96b99d68d27b1fcb75983ef7addc3be6576debbb/chain/blockchain.go#L45-L65

chain/canton/canton_chain.go
gustavogama-cll: I don't know how to implement the same concept for Canton. It's a very peculiar chain family -- for instance, it doesn't have a deployer key, or a signer, or anything equivalent.

So I had to choises: make `ReadOnly` a NOP or always raise "not implemented". So I opted to err on the side of "allowing the hooks to run".

chain/internal/common/chain_metadata.go
gustavogama-cll: import cycle

ecPablo (approved):

PR #963, opened 2026-04-27
feat: implement external library for operations generation

RayXpub:
## Summary

- Introduces an importable tools/operations-gen/generate library package so repos can call generation directly from Go code instead of shelling out to the operations-gen binary. The CLI now delegates to the same library path.
- Fixes generated owner-gated write operations so OnlyOwner is called through an explicit IsAllowedCaller closure, matching the role-gated path and avoiding function signature mismatches.
- Adds RBAC timelock generation coverage to validate role-based operation generation against a realistic fixture.
- Refactors EVM generator unit tests to use testify/require for clearer assertions and cleaner failure handling.
- Added support for a shared EVM `input.gobindings_package` parent path, allowing each contract’s gobindings import to be derived as `<input.gobindings_package>/<version_path>/<package_name>`, while preserving per-contract gobindings_package as an override.


tools/operations-gen/testdata/evm/rbac_timelock.golden.go
```diff
@@ -0,0 +1,73 @@
+// Code generated by operations-gen. DO NOT EDIT.
+
+package rbac_timelock
```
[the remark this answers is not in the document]
RayXpub: Fixed typo in file name

tools/operations-gen/generate/templates/evm/operations.tmpl
```diff
@@ -77,7 +77,9 @@ func NewWrite{{.Name}}(c gobindings.{{$.ContractType}}Interface) *cld_ops.Operat
 		ContractABI:  gobindings.{{$.ContractType}}MetaData.ABI,
 		Contract:     c,
 	{{- if eq .AccessControl "owner"}}
-		IsAllowedCaller: contract.OnlyOwner[gobindings.{{$.ContractType}}Interface, {{.ArgsType}}],
+		IsAllowedCaller: func(c gobindings.{{$.ContractType}}Interface, opts *bind.CallOpts, caller common.Address, args {{.ArgsType}}) (bool, error) {
+			return contract.OnlyOwner(c, opts, caller, args)
```
graham-chainlink, line 81: i noticed `args` is not being used inside `OnlyOwner` ,maybe we could remove it?

tools/operations-gen/README.md
graham-chainlink, line 1: Maybe we could update the readme to include how to use the operations-gen as a package in addition to the CLI?

tools/operations-gen/generate/generate.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if err := handler.Generate(config, tmpl); err != nil {
+		return fmt.Errorf("generate operations: %w", err)
+	}
+
+	return nil
+}
+
+// LoadTemplate loads the code generation template for the given chain family.
+func LoadTemplate(chainFamily string) (*template.Template, error) {
```
graham-chainlink, line 74: Can this function be unexported?

tools/operations-gen/generate/generate.go
```diff
@@ -0,0 +1,101 @@
...
+	content, err := templatesFS.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("template not found at %s: %w", path, err)
+	}
+
+	return template.New("operations").Parse(string(content))
+}
+
+// SupportedFamilies returns a sorted, comma-separated list of supported chain families.
+func SupportedFamilies() string {
```
graham-chainlink: Can this function be unexported too?

tools/operations-gen/generate/generate.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/smartcontractkit/chainlink-deployments-framework/tools/operations-gen/internal/families/evm"
+)
+
+//go:embed templates
+var templatesFS embed.FS
+
+// chainFamilies is the single registration point for all supported chain families.
+var chainFamilies = map[string]core.ChainFamilyHandler{
+	"evm": evm.Handler{},
+}
```
graham-chainlink: Feel free to ignore this comment: i wonder if it is worth introducing a new file call chain.go or handlers.go where all it has is 

```
var chainFamilies = map[string]core.ChainFamilyHandler{
	"evm": evm.Handler{},
}
```
So it is obvious where to hook up a new chain family when time comes?

tools/operations-gen/README.md
graham-chainlink: We can document the 2 entry point for package which is GenerateFile and Generate right?

tools/operations-gen/internal/families/evm/config.go
```diff
@@ -7,11 +7,18 @@ type EvmContractConfig struct {
 	VersionPath       string              `yaml:"version_path,omitempty"` // Optional: override folder path derived from version
 	PackageName       string              `yaml:"package_name,omitempty"` // Optional: override package name
 	OmitDeploy        bool                `yaml:"omit_deploy,omitempty"`  // Optional: skip Deploy operation
-	GobindingsPackage string              `yaml:"gobindings_package"`     // Required: the Go import path of the gobindings package for this contract.
+	GobindingsPackage string              `yaml:"gobindings_package"`     // Optional: override the derived Go import path of the gobindings package for this contract.
```
graham-chainlink: If this is optional, we should include what is the default value here too?

tools/operations-gen/main.go
graham-chainlink, line 1: I think it is more go convention to have the cli under `cmd` eg `cmd/main.go`, thoughts?

tools/operations-gen/generate/templates/evm/operations.tmpl
RayXpub: This is required by the `WriteParams`: https://github.com/smartcontractkit/chainlink-deployments-framework/blob/fc45f9062c6e690c0f80950dc45c915cdc4763e9/chain/evm/operations/contract/write.go#L57

This was not introduced by the refactor, I assume this is mostly for forward compatibility if we introduce an access control hook that may rely on those.

tools/operations-gen/generate/generate.go
RayXpub: This is currently used in `evm_golden_test.go`: https://github.com/smartcontractkit/chainlink-deployments-framework/blob/fc45f9062c6e690c0f80950dc45c915cdc4763e9/tools/operations-gen/internal/families/evm/evm_golden_test.go#L55

tools/operations-gen/internal/families/evm/config.go
RayXpub: This is the per contract path, the default value is still provided by the user but as global config applied to all contracts: https://github.com/smartcontractkit/chainlink-deployments-framework/blob/fc45f9062c6e690c0f80950dc45c915cdc4763e9/tools/operations-gen/internal/families/evm/config.go#L15-L20

tools/operations-gen/generate/templates/evm/operations.tmpl
graham-chainlink: gotcha!

tools/operations-gen/generate/generate.go
graham-chainlink: ah okay, i didn realize the test has a different package name

tools/operations-gen/main.go
RayXpub: Done

tools/operations-gen/generate/generate.go
RayXpub: Yes this one can, updated.

tools/operations-gen/README.md
RayXpub: I updated the README and added library usage section.

tools/operations-gen/main.go
graham-chainlink, line 1: oh you dont need operations-gen under `cmd` since it is already `operations-gen/cmd`

tools/operations-gen/generate/generate.go
RayXpub: Done

tools/operations-gen/main.go
RayXpub: I split it to be more forward compatible if we ever add more commands but makes sense to flatten atm. Done.

tools/operations-gen/generate/generate.go
```diff
@@ -0,0 +1,82 @@
...
+		return fmt.Errorf("generate operations: %w", err)
+	}
+
+	return nil
+}
+
+// LoadTemplate loads the code generation template for the given chain family.
+func LoadTemplate(chainFamily string) (*template.Template, error) {
+	path := fmt.Sprintf("templates/%s/operations.tmpl", chainFamily)
+	content, err := templatesFS.ReadFile(path)
```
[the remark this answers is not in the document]
RayXpub: Refactored to use an injectable logger.

tools/operations-gen/internal/families/evm/contract.go
```diff
@@ -86,6 +83,11 @@ func extractContractInfo(cfg EvmContractConfig, output EvmOutputConfig) (*Contra
 		return nil, err
 	}
 
+	cfg.GobindingsPackage = resolveGobindingsPackage(cfg.GobindingsPackage, input.GobindingsPackage, versionPath, packageName)
+	if cfg.GobindingsPackage == "" {
+		return nil, fmt.Errorf("gobindings_package is required for contract %q; set either contract gobindings_package or input.gobindings_package", cfg.Name)
+	}
+
```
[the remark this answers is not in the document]
RayXpub: Added an override test under `evm_test.go`

tools/operations-gen/cmd/operations-gen/main.go
```diff
@@ -37,7 +38,14 @@ func run(args []string, stdout, stderr io.Writer) int {
 		return 0
 	}
 
-	if err := generate.GenerateFile(*configPath); err != nil {
+	lggr, err := logger.New()
```
graham-chainlink: hmm, personally i think as a CLI, it should not output any logger, just output to std out or std err, however on the library side , we could accept a logger.

So as CLI, no logger is provided , we default to stdout/stderr
In library, we could accept logger. But IMO, I dont think we need any logging except the result at the end where we return to the caller (library), we log to stdout/err for cli mode. Thought?

tools/operations-gen/main.go
RayXpub: I had to unflatten, else the binary name becomes `cmd` instead of `operations-gen`.

tools/operations-gen/main.go
```diff
@@ -0,0 +1,54 @@
+package main
+
+import (
+	"flag"
+	"fmt"
```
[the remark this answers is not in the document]
RayXpub: Updated.

tools/operations-gen/generate/generate.go
```diff
@@ -0,0 +1,102 @@
...
+
+// Option configures generation behavior.
+type Option func(*options)
+
+// WithLogger enables progress logging during generation.
+func WithLogger(lggr logger.Logger) Option {
+	return func(opts *options) {
+		opts.logger = lggr
+	}
+}
```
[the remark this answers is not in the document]
RayXpub: Reverting the logger change, it does not bring much atm and adds quite a bit of overhead.

tools/operations-gen/README.md
```diff
@@ -27,14 +27,64 @@ Print the CLI release metadata:
...
+
+Use `Generate` when the caller has already decoded the config:
+
+```go
+var cfg generate.Config
+if err := yaml.Unmarshal(configBytes, &cfg); err != nil {
+	return err
+}
+cfg.ConfigDir = repoRoot // Base for relative output paths and package loading.
+
```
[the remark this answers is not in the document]
RayXpub: Reverted the logger change.

tools/operations-gen/cmd/operations-gen/main.go
RayXpub: I revert the logger change as discussed.

tools/operations-gen/internal/families/evm/evm.go
```diff
@@ -16,6 +16,13 @@ type Handler struct{}
 // Generate decodes each YAML node as an EvmContractConfig, extracts contract info,
 // and writes a generated operations file for each contract.
 func (h Handler) Generate(config core.Config, tmpl *template.Template) error {
+	var input EvmInputConfig
+	if config.Input.Kind != 0 {
+		if err := config.Input.Decode(&input); err != nil {
+			return fmt.Errorf("failed to decode EVM input config: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
RayXpub: Acceptable for now.

graham-chainlink (dismissed):

tools/operations-gen/internal/families/evm/contract.go
```diff
@@ -113,6 +122,80 @@ func extractContractInfo(cfg EvmContractConfig, output EvmOutputConfig) (*Contra
...
+		// Attempt reading go.mod file
+		goModPath := filepath.Join(current, "go.mod")
+		goMod, readErr := os.ReadFile(goModPath)
+		if readErr == nil {
+			// If there is a go.mod file then we found the root, so we fetch the module path
+			modulePath := modfile.ModulePath(goMod)
+			if modulePath == "" {
+				return "", "", fmt.Errorf("go.mod %q does not define a module path", goModPath)
+			}
+
```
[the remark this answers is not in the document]
RayXpub: Fixed

graham-chainlink (approved):

PR #976, opened 2026-05-14
feat: add cre api keys multiple named secret

ecPablo:
This pull request introduces support for multiple named API keys in the `cre/cli` package, allowing the `CLIRunner` to handle a JSON object of named keys via the `CRE_API_KEY` environment variable. It adds methods to select a key by name, ensures backward compatibility with single-key usage, and provides comprehensive tests for both modes. Mock support for the new interface is also included.

**Support for multiple named API keys:**

* Enhanced `CLIRunner` to accept a JSON object of named API keys via `CRE_API_KEY`, enabling selection of a specific key by name using the new `WithNamedAPIKey` method. This allows workflows to switch between environments or credentials without changing environment variables. [[1]](diffhunk://#diff-73ff3f2caac893a764145fc2bce4bc67f6075239f9a83220e0ba8b0fcf4683d0R26-R50) [[2]](diffhunk://#diff-73ff3f2caac893a764145fc2bce4bc67f6075239f9a83220e0ba8b0fcf4683d0L48-R126) [[3]](diffhunk://#diff-80d026f64febe793cf0db33af0c56a73aac78222777f846e4c1f8a19158445f5R121-R124)
* Added logic to ensure that if multiple named keys are provided, a key must be selected before running commands; otherwise, an error is returned listing available names.

**Testing and compatibility:**

* Added extensive tests for both single-key and named-keys modes, including edge cases, error handling, and ensuring that parent runners are not mutated by key selection.
* Maintained backward compatibility by preserving previous behavior when the API key is not a valid JSON object.

**Mock and interface updates:**

* Updated the `CLIRunner` interface and its mocks to include the new `WithNamedAPIKey` method for use in testing and downstream code. [[1]](diffhunk://#diff-80d026f64febe793cf0db33af0c56a73aac78222777f846e4c1f8a19158445f5R121-R124) [[2]](diffhunk://#diff-93de2268096b1c2158867923a8bdd1f1d9017a3569e7c18e1457a7b53b1d27aaR169-R230)

### DIFF
--- a/cre/cli/runner.go
+++ b/cre/cli/runner.go
@@ -112,7 +112,6 @@ func parseNamedAPIKeys(raw string) (namedAPIKeys, bool) {
 // registered under name. the clone preserves the underlying name->key map so further WithNamedAPIKey
 // calls remain valid.
 func (r cliRunner) WithNamedAPIKey(name string) (fcre.CLIRunner, error) {
-
 	if len(r.apiKeysByName) == 0 {
 		return nil, errors.New("cre cli: runner is not configured with named API keys")
 	}
--- a/cre/cli/runner.go
+++ b/cre/cli/runner.go
@@ -3,10 +3,13 @@ package cli
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
+	"sort"
 	"strings"
 
 	fcre "github.com/smartcontractkit/chainlink-deployments-framework/cre"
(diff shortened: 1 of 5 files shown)


cre/cli/runner.go
```diff
@@ -45,17 +68,62 @@ func NewCLIRunner(binaryPath string, apiKey string, opts ...CLIRunnerOption) *cl
...
+		}
+	}
+
+	return keys, true
+}
+
+// WithNamedAPIKey returns a clone of this runner that uses the API key
+// registered under name. the clone preserves the underlying name->key map so further WithNamedAPIKey
+// calls remain valid.
+func (r *cliRunner) WithNamedAPIKey(name string) (fcre.CLIRunner, error) {
```
graham-chainlink: just a thought, would it be simpler if `r` is not a pointer? Make this method a value receiver, then the clone happen automatically?

cre/cli/runner.go
```diff
@@ -45,17 +68,62 @@ func NewCLIRunner(binaryPath string, apiKey string, opts ...CLIRunnerOption) *cl
 	// formatting (newlines/indentation) during manual durable-pipeline runs.
 	r := &cliRunner{
 		binaryPath: binaryPath,
-		apiKey:     apiKey,
 		Stdout:     os.Stdout,
 		Stderr:     os.Stderr,
 	}
+	if named, ok := parseNamedAPIKeys(apiKey); ok {
```
graham-chainlink: i wonder if it would be better if we dont reuse the `apiKey` field to store the json, and use a new field? This makes this more obvious when debugging and validation? In scenario such as user provide a bad value for json but it is then set as single value for `apiKey` , feels confusing?

cre/cli/runner.go
ecPablo: sure, just changed it here https://github.com/smartcontractkit/chainlink-deployments-framework/pull/976/commits/9ac633340bf19f83986f4753d863a38fb8c20f6b

cre/cli/runner.go
ecPablo: I added a rawApiKey just keep the raw value separate and then the parsed value would vary depending on the if the value is a json or not https://github.com/smartcontractkit/chainlink-deployments-framework/pull/976/commits/9ac633340bf19f83986f4753d863a38fb8c20f6b

graham-chainlink: Ah I think I confused you,  I meant for the constructor NewCLIRunner to have a new param for the json API key or functional option? instead of reusing the existing api key.  It was mainly so if the json is invalid, we don't set the invalid json to the API key field if that make sense ? Is that overkill? The new field added in your latest didn add any value I believe.

ecPablo: Oh I thought you were more interested in having it for tracing/debug reasons. Yeah we coul have the 2 separate keys, I just wanted to abstract this away from clients, otherwise changeset builders will need to know if their secrets were set as a json field or as a single value, The main constraint here is that we want to use the same cre api key secret env var

ChrisAmora (dismissed):

ChrisAmora (approved):

PR #979, opened 2026-05-14
[NONEVM-4520] feat: ContractTypeAndVersion in upf

duck-types:
[NONEVM-4520](https://smartcontract-it.atlassian.net/browse/NONEVM-4520)

Depends on
- https://github.com/smartcontractkit/chainlink-ton/pull/741
- https://github.com/smartcontractkit/chainlink-deployments-framework/pull/981
- https://github.com/smartcontractkit/mcms/pull/751

PR #986, opened 2026-05-18
feat: ContractVersion in OperationMetadata (#979)

duck-types:
feat: ContractVersion in OperationMetadata (#979)
feat: ContractFullyQualifiedName in TON's AdditionalFields
fix: revert ContractTypeAndVersion in OperationMetadata (#979)


experimental/analyzer/decoded_call.go
```diff
@@ -31,30 +31,35 @@ func (d *DecodedCall) String(context *FieldContext) string {
...
 	ct := string(tv.Type)
 	if ct == "" {
 		ct = mcmsTx.ContractType
 	}
 
-	var cv string
-	if tv.Version.Original() != "" {
-		cv = tv.Version.String()
+	// If version wasn't already set from transaction metadata, use datastore version
+	if contractVersion == "" && tv.Version.Original() != "" {
```
gustavogama-cll: I think it should be the opposite: the datastore value should have precedence over the proposal metadata.

@ecPablo @ChrisAmora , what are your thoughts?

experimental/analyzer/upf/upf_test.go
```diff
@@ -231,6 +231,73 @@ func TestUpfConvertTimelockProposalWithTon(t *testing.T) {
...
+		},
+	}
+
+	signers := map[mcmstypes.ChainSelector][]common.Address{
+		mcmstypes.ChainSelector(chainsel.TON_TESTNET.Selector): {
+			common.HexToAddress("0xA5D5B0B844c8f11B61F28AC98BBA84dEA9b80953"),
+		},
+	}
+
+	got, err := UpfConvertProposal(t.Context(), proposalCtx, env, proposal, signers)
```
gustavogama-cll: I'd recommend modifying the test such that the entire contents of the upf file are validated, such as we do in `TestUpfConvertTimelockProposalWithSui`. Or you can simply extend `TestUpfConvertTimelockProposal`, adding a TON operation (it has EVM and Solana already).

gustavogama-cll (commented):
Let's just settle how to handle [this](https://github.com/smartcontractkit/chainlink-deployments-framework/pull/986/changes#r3262092569) (because it affects other chains too) and then we can merge. Though I'd appreciate if you could enhance the UPF test as requested.

experimental/analyzer/decoded_call.go
ecPablo: Yes, it's more likely that datastore has more up-to-date data. For example, I generate proposalA that uses contractVersion 1.0.0, then very shortly after, someone updates datastore to contractVersion 2.0.0. I'd expect the decoded version to show me the 2.0.0 as that's the current on chain state. The only reasons where this could not be true is if we have out-of-cld systems updating the contracts, in that case maybe the proposal might be safer source of truth. But I don't think that's the case here.

experimental/analyzer/upf/upf_test.go
```diff
@@ -231,6 +231,73 @@ func TestUpfConvertTimelockProposalWithTon(t *testing.T) {
...
+		Account: targetAddr,
+	})
+
+	tx, err := mcmstonsdk.NewTransaction(
+		targetAddr,
+		grantRoleData.ToBuilder().ToSlice(),
+		big.NewInt(0),
+		bindings.ShortRBAC,
+		semver.MustParse("1.2.3"),
+		bindings.TypeRBAC,
```
krebernisak: nit: for TON we don't need to input both the `bindings.ShortRBAC` (short) and `bindings.TypeRBAC` (full) as the former can be derived from the latter

experimental/analyzer/ton_analyzer.go
```diff
@@ -37,29 +36,41 @@ func AnalyzeTONTransactions(ctx ProposalContext, chainSelector uint64, txs []typ
...
-	decodedOp, err := decoder.Decode(mcmsTx, contractTypeAndVersion.Type.String())
+	var typeErr string
+	fullyQualifiedName := func() string {
+		var additionalFields ton.AdditionalFields
+		if err := json.Unmarshal(mcmsTx.AdditionalFields, &additionalFields); err != nil {
+			typeErr = fmt.Sprintf("failed to unmarshal TON additional fields: %s", err)
+			return ""
+		}
+
+		fullyQualifiedName := string(additionalFields.ContractTypeFull)
```
krebernisak, line 45: nit: Let's extract into a type/func which we document and allows anybody to produce a FQN

experimental/analyzer/ton_analyzer_test.go
```diff
@@ -242,10 +296,9 @@ func (s *testTONSetup) makeGrantRoleTx(t *testing.T, queryID uint64) types.Trans
 
 func (s *testTONSetup) expectedGrantRoleCall(queryID uint64) *DecodedCall {
 	return &DecodedCall{
-		Address:         s.targetAddr.String(),
-		Method:          string(bindings.TypeRBAC) + "::GrantRole(0x95cd540f)",
-		ContractType:    string(bindings.TypeRBAC),
-		ContractVersion: "0.0.0",
+		Address:      s.targetAddr.String(),
+		Method:       string(bindings.TypeRBAC) + "::GrantRole(0x95cd540f)",
```
krebernisak: nit: this DecodedCall should also contain a contract version, maybe as part of the `Method` by using a `FQN` vs `ContractTypeFull` (missing version)

experimental/analyzer/decoded_call.go
krebernisak: Ok, let's keep system context (DS) to have priority. Override can be made with updating the DS directly with the proposal PR (if something is off).

Finally, we should also be able to interact with ad-hoc contracts which are not stored, and for that cases the proposal input ref (type, version) should resolve properly.

krebernisak (dismissed):

experimental/analyzer/ton_analyzer_test.go
duck-types: This was already done but it was not covered by test. Added a case to confirm this

experimental/analyzer/upf/upf_test.go
duck-types: Not if the full type is a trait type. Short type is only used for the contract type as stored in the DS.

experimental/analyzer/ton_analyzer.go
```diff
@@ -29,37 +28,43 @@ func AnalyzeTONTransactions(ctx ProposalContext, chainSelector uint64, txs []typ
...
+		// If ContractVersion is provided, append it to the fully qualified name to ensure the decoder uses the correct version.
+		// If it is skipped, the decoder will use the latest version available for the contract type.
+		// Note: we don't use contractVersion from resolveContractInfo because that only represents the short type used by the datastore.
+		if mcmsTx.ContractVersion != nil {
+			fullyQualifiedName += "@" + mcmsTx.ContractVersion.String()
+		}
```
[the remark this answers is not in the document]
duck-types: We don't want this in TON because we only want to select a version explicitly via TX params

gustavogama-cll (approved):

PR #1006, opened 2026-05-26
[CLD-2461]: feat(deployment): introduce mcms registry

graham-chainlink:
As part of integrating the utils from Chainlink CCIP tooling API, we are bringing in the [MCMSInput](https://github.com/smartcontractkit/chainlink-ccip/blob/58d2d8467c93ef8db4a434da71062189fd74312c/deployment/utils/mcms) and [MCMSRegistry](https://github.com/smartcontractkit/chainlink-ccip/blob/58d2d8467c93ef8db4a434da71062189fd74312c/deployment/utils/changesets/output.go#L31) plus some code changes, this will be used by the incoming ChangesetOutput.

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-2461

### DIFF
--- a/deployment/mcms_registry.go
+++ b/deployment/mcms_registry.go
@@ -1,0 +1,147 @@
+package deployment
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
+
+	mcmstypes "github.com/smartcontractkit/mcms/types"
+)
+
+var (
+	// ErrInvalidMCMSTimelockProposalInput indicates MCMSTimelockProposalInput failed validation.
+	ErrInvalidMCMSTimelockProposalInput = errors.New("invalid MCMS timelock proposal input")
+	// ErrDuplicateMCMSReader indicates a reader is already registered for a chain family.
+	ErrDuplicateMCMSReader = errors.New("MCMS reader already registered for chain family")
+	// ErrEmptyChainFamily indicates Register was called with an empty chain family.
+	ErrEmptyChainFamily = errors.New("chain family must not be empty")
+	// ErrNilMCMSReader indicates Register was called with a nil reader.
+	ErrNilMCMSReader = errors.New("MCMS reader must not be nil")
+)
+
+// MCMSTimelockProposalInput is an input for creating an MCMS timelock proposal.
+type MCMSTimelockProposalInput struct {
+	// OverridePreviousRoot indicates whether to override the root of the MCMS contract.
+	OverridePreviousRoot bool
+	// ValidUntil is a Unix timestamp indicating when the proposal expires.
+	// Root can't be set or executed after this time.
+	ValidUntil uint32
+	// TimelockDelay is the amount of time each operation in the proposal must wait before it can be executed.
+	TimelockDelay mcmstypes.Duration
+	// TimelockAction is the action to perform on the timelock contract (schedule, bypass, or cancel).
+	TimelockAction mcmstypes.TimelockAction
+	// Qualifier is a string used to qualify the MCMS + Timelock contract addresses.
+	Qualifier string
+	// Description is a human-readable description of the proposal.
+	Description string
+}
+
+// minValidUntil returns the earliest ValidUntil timestamp accepted by Validate.
+func minValidUntil(now time.Time) uint32 {
+	return uint32(now.Add(10 * time.Minute).UTC().Unix()) //nolint:gosec // Unix timestamp fits uint32 until year 2106
+}
+
+// Validate validates the MCMSTimelockProposalInput.
+func (c *MCMSTimelockProposalInput) Validate() error {
+	return c.validateAt(time.Now())
+}
+
+func (c *MCMSTimelockProposalInput) validateAt(now time.Time) error {
+	if c.TimelockAction != mcmstypes.TimelockActionSchedule &&
+		c.TimelockAction != mcmstypes.TimelockActionBypass &&
+		c.TimelockAction != mcmstypes.TimelockActionCancel {
+		return fmt.Errorf("%w: invalid timelock action %q", ErrInvalidMCMSTimelockProposalInput, c.TimelockAction)
+	}
+	if c.TimelockDelay.Duration < 0 {
+		return fmt.Errorf("%w: timelock delay must not be negative", ErrInvalidMCMSTimelockProposalInput)
+	}
+	if c.TimelockAction == mcmstypes.TimelockActionSchedule && c.TimelockDelay.Duration <= 0 {
+		return fmt.Errorf("%w: timelock delay must be positive for schedule action", ErrInvalidMCMSTimelockProposalInput)
+	}
+	if c.ValidUntil == 0 {
+		return fmt.Errorf("%w: valid until must be set", ErrInvalidMCMSTimelockProposalInput)
+	}
+	// valid until must be far enough in the future to account for clock drift and proposal creation delay.
+	if c.ValidUntil < minValidUntil(now) {
+		return fmt.Errorf("%w: valid until must be at least 10 minutes in the future", ErrInvalidMCMSTimelockProposalInput)
+	}
+
+	return nil
+}
+
+// MCMSReader is an interface for reading MCMS state from a chain type.
+type MCMSReader interface {
+	// GetChainMetadata returns the chain metadata for a given MCMS input.
+	// Each chain family defines its own implementation of this method.
+	GetChainMetadata(e Environment, chainSelector uint64, input MCMSTimelockProposalInput) (mcmstypes.ChainMetadata, error)
+	// GetTimelockRef returns the timelock contract address reference for a given MCMS input.
+	GetTimelockRef(e Environment, chainSelector uint64, input MCMSTimelockProposalInput) (datastore.AddressRef, error)
+	// GetMCMSRef returns the MCMS contract address reference for a given MCMS input.
+	GetMCMSRef(e Environment, chainSelector uint64, input MCMSTimelockProposalInput) (datastore.AddressRef, error)
+}
+
+// MCMSReaderRegistry maintains a registry of MCMS readers.
+type MCMSReaderRegistry struct {
+	mu sync.RWMutex
+	m  map[string]MCMSReader
+}
+
+func newMCMSReaderRegistry() *MCMSReaderRegistry {
+	return &MCMSReaderRegistry{
+		m: make(map[string]MCMSReader),
+	}
+}
+
+var (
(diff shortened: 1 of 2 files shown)


deployment/mcms_registry.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	if c.ValidUntil == 0 {
+		return fmt.Errorf("%w: valid until must be set", ErrInvalidMCMSTimelockProposalInput)
+	}
+	// valid until must be far enough in the future to account for clock drift and proposal creation delay.
+	if c.ValidUntil < minValidUntil(now) {
+		return fmt.Errorf("%w: valid until must be at least 10 minutes in the future", ErrInvalidMCMSTimelockProposalInput)
+	}
+
+	return nil
```
graham-chainlink, line 73: @ecPablo and @gustavogama-cll does my validations here make sense logically for MCMS?

deployment/mcms_registry.go
```diff
@@ -0,0 +1,147 @@
...
+	OverridePreviousRoot bool
+	// ValidUntil is a Unix timestamp indicating when the proposal expires.
+	// Root can't be set or executed after this time.
+	ValidUntil uint32
+	// TimelockDelay is the amount of time each operation in the proposal must wait before it can be executed.
+	TimelockDelay mcms_types.Duration
+	// TimelockAction is the action to perform on the timelock contract (schedule, bypass, or cancel).
+	TimelockAction mcms_types.TimelockAction
+	// Qualifier is a string used to qualify the MCMS + Timelock contract addresses.
+	Qualifier string
```
gustavogama-cll, line 38: noaction: I fear we might be following CCIP's solution too closely and then end up alienating the other teams. For instance, we already got feedback from payments and treasury that they'd lie to to have a set of MCM and timelock contracts that doesn't follow the current standards.

deployment/mcms_registry.go
```diff
@@ -0,0 +1,147 @@
...
+	// Description is a human-readable description of the proposal.
+	Description string
+}
+
+// minValidUntil returns the earliest ValidUntil timestamp accepted by Validate.
+func minValidUntil(now time.Time) uint32 {
+	return uint32(now.Add(10 * time.Minute).UTC().Unix()) //nolint:gosec // Unix timestamp fits uint32 until year 2106
```
gustavogama-cll, line 45: 10m? Where did this come from?

deployment/mcms_registry.go
gustavogama-cll: Yes, kind of. Not sure we need to put a lower bound on valid until but I guess it doesn't hurt.

What would be *really* good to add (though I don't think is possible) is to validate if the TimelockDelay value is equal or greater than the onchain value.
Maybe we can add that check in the CLD engine somewhere.

deployment/mcms_registry.go
```diff
@@ -0,0 +1,147 @@
...
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
+
+	mcms_types "github.com/smartcontractkit/mcms/types"
```
jkongie: *nit* no underscores for consistency

```suggestion
	mcmstypes "github.com/smartcontractkit/mcms/types"
```

deployment/mcms_registry.go
```diff
@@ -0,0 +1,147 @@
...
+func GetMCMSReaderRegistry() *MCMSReaderRegistry {
+	once.Do(func() {
+		singletonRegistry = newMCMSReaderRegistry()
+	})
+
+	return singletonRegistry
+}
+
+// RegisterMCMSReader registers an MCMSReader for a specific chain family.
+func (r *MCMSReaderRegistry) RegisterMCMSReader(chainFamily string, reader MCMSReader) error {
```
jkongie:
```suggestion
func (r *MCMSReaderRegistry) Register(chainFamily string, reader MCMSReader) error {
```

deployment/mcms_registry.go
```diff
@@ -0,0 +1,147 @@
...
+	chainFamily = strings.TrimSpace(chainFamily)
+	if chainFamily == "" {
```
gustavogama-cll, line 118: shouldn't we check if this chain family exists in chain-selectors?

deployment/mcms_registry.go
jkongie: Do you need to provide the chainFamily? I think you could fetch this from the `MCMSReader`?

deployment/mcms_registry.go
```diff
@@ -0,0 +1,147 @@
...
+	if _, exists := r.m[chainFamily]; exists {
+		return fmt.Errorf("%w: %q", ErrDuplicateMCMSReader, chainFamily)
+	}
+	r.m[chainFamily] = reader
+
+	return nil
+}
+
+// GetMCMSReader retrieves an MCMSReader for a specific chain family.
+func (r *MCMSReaderRegistry) GetMCMSReader(chainFamily string) (MCMSReader, bool) {
```
jkongie:
```suggestion
func (r *MCMSReaderRegistry) Get(chainFamily string) (MCMSReader, bool) {
```

gustavogama-cll (dismissed):

deployment/mcms_registry.go
```diff
@@ -0,0 +1,147 @@
...
+
+var (
+	singletonRegistry *MCMSReaderRegistry
+	once              sync.Once
+)
+
+// GetMCMSReaderRegistry returns the global singleton MCMS reader registry.
+// The first call creates the registry; subsequent calls return the same pointer.
+// This is the recommended way to get the registry, as it ensures a single instance is created and shared.
+func GetMCMSReaderRegistry() *MCMSReaderRegistry {
```
jkongie, line 107: Where do people call this?

deployment/mcms_registry.go
```diff
@@ -0,0 +1,147 @@
...
+	if c.TimelockAction == mcms_types.TimelockActionSchedule && c.TimelockDelay.Duration <= 0 {
+		return fmt.Errorf("%w: timelock delay must be positive for schedule action", ErrInvalidMCMSTimelockProposalInput)
+	}
+	if c.ValidUntil == 0 {
+		return fmt.Errorf("%w: valid until must be set", ErrInvalidMCMSTimelockProposalInput)
+	}
+	// valid until must be far enough in the future to account for clock drift and proposal creation delay.
+	if c.ValidUntil < minValidUntil(now) {
+		return fmt.Errorf("%w: valid until must be at least 10 minutes in the future", ErrInvalidMCMSTimelockProposalInput)
+	}
```
jkongie, line 71: Is this really for us to define?

deployment/mcms_registry.go
graham-chainlink: This was from CCIP, but i wanted to check if this is a CCIP thing or in general good to enforce?

deployment/mcms_registry.go
graham-chainlink: Good questions, i wanted to check with the @ecPablo  and @gustavogama-cll  on whether this is a CCIP specific logic or in MCMS general

deployment/mcms_registry.go
gustavogama-cll: I don't think it's a _bad_ idea to add. But, on the other hand, 10m is a bit arbitrary. On the other hand, making it configurable per domain seem overkill... I don't know, maybe we just leave it as is. As I said earlier, I don't think it _hurts_ to have this check.

ecPablo (dismissed):

deployment/mcms_registry.go
```diff
@@ -0,0 +1,147 @@
...
+
+	return nil
+}
+
+// MCMSReader is an interface for reading MCMS state from a chain type.
+type MCMSReader interface {
```
ecPablo, line 77: nit(no action): I can't think of a better name, but when looking at `MCMSReader` first thing I thought was that it was simiar to the `sdk.Inspector` interface from mcmslib. Wonder if there's a way we can make this more explicit that is a reader related more to the datastore data and not on chain data for mcms

deployment/mcms_registry.go
graham-chainlink: Can you enlighten me on this? Do you know what other teams may prefer or how i can make it more general? Since qualifier is a string, is that enough? Or is qualifier a concept only familiar to CCIP?

deployment/mcms_registry.go
graham-chainlink: okay sounds good, will leave it as it is and if needed i can make this customisable in future

deployment/mcms_registry.go
graham-chainlink: > Do you need to provide the chainFamily? I think you could fetch this from the MCMSReader?

I think we need to have the chainFamily to serve as the lookup key, we cant call any of the methods in MCMSReader at registration time as we dont have the inputs during registration.
We could however introduce a new method `ChainFamily()` under `MCMSReader` interface but then we are adding more to the interface.

deployment/mcms_registry.go
graham-chainlink: This will be used in the ChangesetOutputBuilder which i will add in another PR, and consumers will use this in their changeset to register MCMS Reader

deployment/mcms_registry.go
graham-chainlink: Discussed with Gustavo [here](https://github.com/smartcontractkit/chainlink-deployments-framework/pull/1006#discussion_r3304995702), looks like we are _okay_ to include it

deployment/mcms_registry.go
graham-chainlink: Unfortunately the chain-selectors package doesnt offer a way to validate only chain family, however it is not as bad as i thought as the key itself is just a pure lookup string key from the perspective of the MCMS registry, so it could be anything.

deployment/mcms_registry.go
graham-chainlink: I see where you are coming from, i could go with `MCMSTimelockProposalResolver` as that is more specific but it may limit the type of data we add to the interface. However keeping this name would be familiar for the CCIP devs as they have been using it.

deployment/mcms_registry.go
gustavogama-cll: The question is: can/should you use the qualifier to group the MCMS contracts? What we heard from those teams is that they wanted to have multiple timelocks managed by the same MCM contracts and/or vice-versa -- multiple MCMs managing the same timelock (with the same role).

A qualifier works well as an identifier for an exclusive set of contracts. I don't know what we'd need to handle those other setups.

gustavogama-cll (approved):

graham-chainlink:
Merging this in first , any further comments i am happy to address in next PR

PR #1007, opened 2026-05-26
[CLD-2461]: feat(datastore): new WriteMetadata util

graham-chainlink:
Chainlink CCIP tooling api provides [WriteMetadataToDatastore util](https://github.com/smartcontractkit/chainlink-ccip/blob/298ed1c38d5cda61a688beb2494a96a5409daf3d/deployment/utils/sequences/sequences.go#L76), this is heavily used in their changets.

I have integrated this into the datastore package with the MutableDatastore and added the ability to add addresses as well as that is a common pattern.

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-2461

### DIFF
--- a/datastore/write_metadata.go
+++ b/datastore/write_metadata.go
@@ -1,0 +1,93 @@
+package datastore
+
+import "fmt"
+
+// MetadataBundle is address refs plus contract, chain, and env metadata.
+type MetadataBundle struct {
+	// Addresses are the contract addresses that were deployed.
+	Addresses []AddressRef
+	// Contracts defines any metadata pertaining to contracts that were deployed.
+	Contracts []ContractMetadata
+	// Chains defines metadata pertaining to chains that were operated against.
+	Chains []ChainMetadata
+	// Env defines any metadata pertaining to the environment that was deployed to.
+	Env *EnvMetadata
+}
+
+type addressRefWriteMode int
+
+const (
+	addressRefWriteAdd addressRefWriteMode = iota
+	addressRefWriteUpsert
+)
+
+type writeMetadataConfig struct {
+	addressRefMode addressRefWriteMode
+}
+
+// WriteMetadataOption configures WriteMetadataToDatastore.
+type WriteMetadataOption func(*writeMetadataConfig)
+
+// WithUpsertAddressRefs writes address refs with Upsert instead of the default Add.
+// Upsert replaces the full record for an existing key (chain, type, version, qualifier),
+// including Address.
+func WithUpsertAddressRefs() WriteMetadataOption {
+	return func(cfg *writeMetadataConfig) {
+		cfg.addressRefMode = addressRefWriteUpsert
+	}
+}
+
+func applyWriteMetadataOptions(opts []WriteMetadataOption) writeMetadataConfig {
+	cfg := writeMetadataConfig{addressRefMode: addressRefWriteAdd}
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+
+	return cfg
+}
+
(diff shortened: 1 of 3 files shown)


datastore/write_metadata.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// WriteMetadataToDatastore adds address refs and upserts contract and chain metadata and sets env metadata.
+func WriteMetadataToDatastore(ds MutableDataStore, bundle MetadataBundle) error {
+	for _, ref := range bundle.Addresses {
+		if err := ds.Addresses().Add(ref); err != nil {
+			return fmt.Errorf("failed to add %s %v at %s on chain %d to datastore: %w",
+				ref.Type, ref.Version, ref.Address, ref.ChainSelector, err)
+		}
+	}
```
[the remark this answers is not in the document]
graham-chainlink: existing issue, need to fix properly

datastore/write_metadata.go
```diff
@@ -0,0 +1,44 @@
...
+	// Contracts defines any metadata pertaining to contracts that were deployed.
+	Contracts []ContractMetadata
+	// Chain defines any metadata pertaining to the chain that was operated against.
+	Chain *ChainMetadata
+	// Env defines any metadata pertaining to the environment that was deployed to.
+	Env *EnvMetadata
+}
+
+// WriteMetadataToDatastore adds address refs and upserts contract and chain metadata and sets env metadata.
+func WriteMetadataToDatastore(ds MutableDataStore, bundle MetadataBundle) error {
```
graham-chainlink: Instead of extending the interface of `MutableDataStore` to include `WriteMetadata`, i just introduce a helper function here so any variable of type MutableDataStore can use this

datastore/write_metadata.go
```diff
@@ -0,0 +1,44 @@
...
+	for _, ref := range bundle.Addresses {
+		if err := ds.Addresses().Add(ref); err != nil {
```
giogam: Why not `Upsert`? Using `Add` would fail if, for example, someone called `WriteMetadata` intending to update an existing `AddressRef`

datastore/write_metadata.go
```diff
@@ -0,0 +1,44 @@
...
+import "fmt"
+
+// MetadataBundle is address refs plus contract, chain, and env metadata.
+type MetadataBundle struct {
+	// Addresses are the contract addresses that were deployed.
+	Addresses []AddressRef
+	// Contracts defines any metadata pertaining to contracts that were deployed.
+	Contracts []ContractMetadata
+	// Chain defines any metadata pertaining to the chain that was operated against.
+	Chain *ChainMetadata
```
giogam: I get why Chain isn't a slice here, but as we generalize this CCIP-specific logic into a broadly available API, could this become a limitation? It's technically possible to have multiple ChainMetadata records per domain/environment, and other users might need that.

graham-chainlink: okay, good pickup, let me make chain into a slice

datastore/write_metadata.go
graham-chainlink: Yeah this was one of my question, by looking at the usage pattern , so [far all the cases](https://github.com/smartcontractkit/chainlink-ccip/blob/e8daba118223ff7f28627eb0b7c2e672975f3e48/deployment/v2_0_0/changesets/deploy_chain_contracts.go#L209) in chainlink-ccip is that `Add` is being used. However i cant predict future usage pattern, so i will add an option to allow user to select which mode then want

giogam (approved):

PR #1016, opened 2026-05-28
feat(operations): introduce WithIdempotencyKey

graham-chainlink:
Operation API cache/report reuse is based on the definition + input of the operation,  sometimes we have use cases where the input is the same but the dependency is not the same and user would like the operation for it to be executed instead of being reused with the existing cache/reports. Eg when the chain client passed into dep can be for different chains but implements the same interface.

Because of this, users have made [workarounds](https://github.com/smartcontractkit/payments/blob/5462c054fd879a0da77fffd68fa7b0c069d36d2d/changeset/shared/operations/access_control.go#L180-L185) by passing chain-selectors or custom uuid into the input purely for cache breaking reasons.

This change introduce an extra hash key that user can provide as a differentiator. The API will also consider this new key  as part of the hash key for reusing previous successful results. By providing this key, user avoid having to introduce extra placeholder param in the input struct.

```go
result, err = operations.ExecuteOperation(bundle, op, deps, input, operations.WithIdempotencyKey[InputType, DepsType](extra-key))
```

This change originated from discussing with Rayene and Rens.

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-2579

### DIFF
--- a/operations/execute.go
+++ b/operations/execute.go
@@ -15,6 +15,8 @@ type ExecuteConfig[IN, DEP any] struct {
 	retryConfig RetryConfig[IN, DEP]
 	// forceExecute controls whether execution should skip execution when previous successful report is found (set by WithForceExecute).
 	forceExecute bool
+	// idempotencyKey scopes report reuse beyond operation definition and input (set by WithIdempotencyKey).
+	idempotencyKey string
 }
 
 type ExecuteOption[IN, DEP any] func(*ExecuteConfig[IN, DEP])
@@ -94,6 +96,15 @@ func WithForceExecute[IN, DEP any]() ExecuteOption[IN, DEP] {
 	}
 }
 
+// WithIdempotencyKey is an ExecuteOption that adds an extra component to the idempotency hash.
+// The hash will then be built from the operation definition, input, and this key.
+// Use it when the same operation input can legitimately produce different results, so a later run should not reuse an earlier result.
+func WithIdempotencyKey[IN, DEP any](idempotencyKey string) ExecuteOption[IN, DEP] {
+	return func(c *ExecuteConfig[IN, DEP]) {
+		c.idempotencyKey = idempotencyKey
+	}
+}
+
 // WithOperationNRetry is an ExecuteOperationNOption that enables the default retry for each run in ExecuteOperationN.
 func WithOperationNRetry[IN, DEP any]() ExecuteOperationNOption[IN, DEP] {
 	return func(c *ExecuteOperationNConfig[IN, DEP]) {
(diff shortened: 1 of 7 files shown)


operations/hashing.go
```diff
@@ -9,24 +9,37 @@ import (
 	"sync"
 )
 
-func constructUniqueHashFrom(hashCache *sync.Map, def Definition, input any) (string, error) {
-	// convert def and input into string to be used as cachekey
-	// as input can be any type, some types cannot be used as cache argument for sync.map
-	// converting to string ensure argument is always valid
-
+func buildUniqueHashCacheKey(def Definition, input any, idempotencyKey string) ([]byte, error) {
```
graham-chainlink, line 12: A small refactor to extra the cosntruction logic of hash key into `buildUniqueHashCacheKey` from `constructUniqueHashFrom` so it can be reused in the unit test for consistency

operations/execute.go
```diff
@@ -94,6 +96,15 @@ func WithForceExecute[IN, DEP any]() ExecuteOption[IN, DEP] {
 	}
 }
 
+// WithIdempotencyKey is an ExecuteOption that adds an extra component to the idempotency hash.
+// The hash will then be built from the operation definition, input, and this key.
+// Use it when the same operation input can legitimately produce different results, so a later run should not reuse an earlier result.
```
DimitriosNaikopoulos, line 101: Wouldn't that be the same with removing the operations file from the PR? 🤔

DimitriosNaikopoulos: Oh nvm I think I get it now 🤔 This is when the inputs are exactly the same but the changeset if is re-executed may produce different result? Not sure how that is possible maybe there is a bit of extra data that we get from somewhere else like on-chain because else it sound a bit unpredicatable 🤔

DimitriosNaikopoulos (approved):

ajaskolski (approved):

jkongie (changes requested):
How do re-runs affect the reports?

Will you have many dead operations in there when you re-run?

If you were to visualize the report as a tree, would you find it hard to understand what was actually executed for the successful run?

graham-chainlink:
It wont be any different to the existing behaviour, the new addition is that now the user can provide their own unique key to be included when generating the operation hash.

on rerun, the same user provided idempotency key of the same operation will return the same result for the same operation.

> Will you have many dead operations in there when you re-run?

Provided user do not provide a randomly generated idempotency key on every rerun, I dont see a reason why this will introduce more dead operations as compared to now? 

 > If you were to visualize the report as a tree, would you find it hard to understand what was actually executed for the successful run?

Hmm i dont see how it would be harder than now? Idempotency key is just another additional key we consider for hashing on top of operation definition and input.

operations/execute.go
graham-chainlink: > Wouldn't that be the same with removing the operations file from the PR? 🤔

we could always just delete the operations reports, but the reports is also there for audit trail reason, so even without the caching, we wan to keep the reports.

> This is when the inputs are exactly the same but the changeset if is re-executed may produce different result

Almost, it is inside the changeset when multiple of the same operation is being re-executed, example would be fetching on chain data before and after it is being updated in the same changeset.

jkongie (approved):

PR #1019, opened 2026-06-02
[CLD-2622]: feat(operations): introduce WithSequenceIdempotencyKey

graham-chainlink:
Similar to WithIdempotencyKey for ExecuteOperation, we want to introduce the same option for Sequence as well.

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-2622

### DIFF
--- a/operations/execute.go
+++ b/operations/execute.go
@@ -21,6 +21,15 @@ type ExecuteConfig[IN, DEP any] struct {
 
 type ExecuteOption[IN, DEP any] func(*ExecuteConfig[IN, DEP])
 
+// ExecuteSequenceConfig holds options for ExecuteSequence.
+type ExecuteSequenceConfig[IN, DEP any] struct {
+	// idempotencyKey scopes report reuse beyond sequence definition and input (set by WithSequenceIdempotencyKey).
+	idempotencyKey string
+}
+
+// ExecuteSequenceOption configures ExecuteSequence.
+type ExecuteSequenceOption[IN, DEP any] func(*ExecuteSequenceConfig[IN, DEP])
+
 // ExecuteOperationNConfig holds options for ExecuteOperationN.
 type ExecuteOperationNConfig[IN, DEP any] struct {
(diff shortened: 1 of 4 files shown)


operations/execute.go
```diff
@@ -98,13 +107,20 @@ func WithForceExecute[IN, DEP any]() ExecuteOption[IN, DEP] {
...
-// Use it when the same operation input can legitimately produce different results, so a later run should not reuse an earlier result.
+// Use it when the same input can legitimately produce different results, so a later run should not reuse an earlier result.
 func WithIdempotencyKey[IN, DEP any](idempotencyKey string) ExecuteOption[IN, DEP] {
 	return func(c *ExecuteConfig[IN, DEP]) {
 		c.idempotencyKey = idempotencyKey
 	}
 }
 
+// WithSequenceIdempotencyKey is an ExecuteSequenceOption with the same semantics as WithIdempotencyKey.
+func WithSequenceIdempotencyKey[IN, DEP any](idempotencyKey string) ExecuteSequenceOption[IN, DEP] {
```
DimitriosNaikopoulos, line 118: nit/question: Should `WithIdempotencyKey` automatically add the key to any sequences as well? With this approach, I think the user needs to know if the changeset is using sequences or not in order to set the IdempotencyKey and some times operators are mostly consumers of changesets with only high level knowledge of the inner workings of a changeset 🤔

DimitriosNaikopoulos (approved):
Approving but I have a question if that could be combined with the previous function 🤔

operations/execute.go
graham-chainlink: the `WithIdempotencyKey` add the key to the operation and `WithSequenceIdempotencyKey` add the key to the sequence, so the changeset writer will need to know about this as they are the one passing in the idempotency key. We dont set the `IdempotencyKey` on the changeset. Consumers of changeset wont need to know about.

I hope this make sense?

PR #1027, opened 2026-06-04
feat(datastore): new query helpers

graham-chainlink:
Inspired by the regular query methods in CCIP tooling api [here](https://github.com/smartcontractkit/chainlink-ccip/blob/6eee37eef754db75bc1aa9baa0be1e3a21730f7d/deployment/utils/datastore/datastore.go) , i integrated the most used utils plus some changes and cleanup

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-2465

### DIFF
--- a/datastore/query.go
+++ b/datastore/query.go
@@ -1,0 +1,114 @@
+package datastore
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+var (
+	ErrAddressRefQueryEmpty     = errors.New("address ref query is empty")
+	ErrAddressRefQueryNoMatch   = errors.New("no address ref matched query")
+	ErrAddressRefQueryAmbiguous = errors.New("multiple address refs matched query")
+	ErrAddressRefFormatFailed   = errors.New("failed to format address ref")
+)
+
+// FormatFn formats a resolved AddressRef into type T (for example a chain-native address).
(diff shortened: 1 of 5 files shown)


datastore/query.go
```diff
@@ -0,0 +1,119 @@
...
+// Version when multiple versions of the same contract type exist on a chain.
+func FindUniqueRef(store AddressRefStore, ref AddressRef) (AddressRef, error) {
+	return findUniqueRef(store, ref)
+}
+
+// FindAndFormatRef resolves a unique AddressRef via [FindUniqueRef] and formats it with format.
+// format must be non-nil.
+func FindAndFormatRef[T any](store AddressRefStore, ref AddressRef, format FormatFn[T]) (T, error) {
+	var empty T
+	if format == nil {
```
giogam, line 39: Just a thought: why not default to `FullRef` instead of returning an error when no formatting function is provided? It would likely be a breaking change, so I assume we can’t change the behavior at this point.

graham-chainlink: nah , we dont care about breaking change since this is new , ccip users can decide if they want to move over or not.

For ur suggestions, since we also have `FindUniqueRef` , they should use that instead of `FindAndFormatRef` if they want the `addressRef` as the return value, but this means i can actually remove `FullRef` since we dont need that anymore , will do that instead!

giogam (approved):

PR #1041, opened 2026-06-10
feat(utils): introduce ExecuteOnChainSequenceAndMerge

graham-chainlink:
Introduced ExecuteOnChainSequenceAndMerge util which originated from ccip tooling api [here](https://github.com/smartcontractkit/chainlink-ccip/blob/298ed1c38d5cda61a688beb2494a96a5409daf3d/deployment/utils/sequences/sequences.go#L46)

This utils allows user to execute multiple sequence and accumulate their results into an object call OnChainOutput

```go
import (
    "github.com/smartcontractkit/chainlink-deployments-framework/changeset/sequenceutils"
)
// Run multiple sequences and accumulate into one OnChainOutput.
var agg sequenceutils.OnChainOutput
agg, err = sequenceutils.ExecuteOnChainSequenceAndMerge(
    env.OperationsBundle, deps, deploySeq, deployInput, agg,
)
if err != nil {
    return agg, err
}
agg, err = sequenceutils.ExecuteOnChainSequenceAndMerge(
    env.OperationsBundle, deps, configureSeq, configureInput, agg,
)
if err != nil {
    return agg, err // prior merges preserved in agg
}

```

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-2464

### DIFF
--- a/changeset/sequenceutils/merge.go
+++ b/changeset/sequenceutils/merge.go
@@ -1,0 +1,56 @@
+package sequenceutils
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
+	"github.com/smartcontractkit/chainlink-deployments-framework/deployment"
+	"github.com/smartcontractkit/chainlink-deployments-framework/operations"
+)
+
+// ExecuteOnChainSequenceAndMerge executes a sequence and merges the output into the given OnChainOutput.
+// On error, the accumulated agg is returned unchanged.
+//
+// Env metadata is a single record per deployment. If both agg and the sequence output set env metadata
+// with different values, merge fails with deployment.ErrInvalidConfig rather than silently overwriting.
+// Re-merging the same env (shared pointer or equal Metadata value) is allowed.
+func ExecuteOnChainSequenceAndMerge[IN any, DEP any](
+	b operations.Bundle,
+	deps DEP,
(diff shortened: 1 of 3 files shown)


changeset/sequenceutils/merge.go
```diff
@@ -0,0 +1,56 @@
...
+// On error, the accumulated agg is returned unchanged.
+//
+// Env metadata is a single record per deployment. If both agg and the sequence output set env metadata
+// with different values, merge fails with deployment.ErrInvalidConfig rather than silently overwriting.
+// Re-merging the same env (shared pointer or equal Metadata value) is allowed.
+func ExecuteOnChainSequenceAndMerge[IN any, DEP any](
+	b operations.Bundle,
+	deps DEP,
+	seq *operations.Sequence[IN, OnChainOutput, DEP],
+	input IN,
```
jkongie, line 22: I'm finding the ordering of arguments here unintiutive (seq, deps, input would be better to me) but I assume that this keeps it backwards compatible.

Not a blocker though

jkongie (approved):

changeset/sequenceutils/merge.go
graham-chainlink: ah yes, the sequence generic type ordering has always been like this 
```go
type Sequence[IN, OUT, DEP any] struct {
	def     Definition
	handler SequenceHandler[IN, OUT, DEP]
}

```

PR #1044, opened 2026-06-11
fix(datastore): preserve staged deletes across Merge composition

giogam:
`MemoryDataStore.Merge` propagated `src.DeletedRemoteKeys` by calling `dst.<Store>.Delete(key)`, which removed the record from `dst.Records` but never appended the key to `dst.DeletedRemoteKeys`. Chained Merges (operation and changeset composition) consumed the prior step's staged deletes at every hop; both the JSON file and catalog persistence backends drive remote deletes off DeletedRemoteKeys, so any delete that passed through more than one Merge silently never reached the remote.

Each of the three propagation blocks (`AddressRef`, `ChainMetadata`, `ContractMetadata`) now calls `dst.<Store>.RemoteDelete(key)` followed by `dst.<Store>.Delete(key)` with the per-store `NotFound` sentinel tolerated. Delete intent now survives chained Merges, and Merge is idempotent on repeated calls with the same source.

Behavior change: Merge previously returned Err<X>NotFound when the source staged a delete for a key not present in `dst.Records`; it now succeeds and stages the delete on dst. This aligns observable behavior with the RemoteDelete docstring, which already advertised staging deletes for records not held locally. Three existing tests in `memory_datastore_test.go` are inverted accordingly; new tests cover chained-merge preservation, idempotency, and the precedence of source-live records over destination-staged deletes.

### DIFF
--- a/datastore/memory_datastore.go
+++ b/datastore/memory_datastore.go
@@ -62,7 +62,28 @@ func (s *MemoryDataStore) WriteMetadata(bundle MetadataBundle, opts ...WriteMeta
 	return WriteMetadataToDataStore(s, bundle, opts...)
 }
 
-// Merge merges the given mutable data store into the current MemoryDataStore.
+// Merge applies records and staged deletions from other onto this MemoryDataStore.
+//
+// Staged deletions (other.<Store>.DeletedRemoteKeys) are propagated by appending the
+// key to s.<Store>.DeletedRemoteKeys via RemoteDelete and then removing the record
+// from s.<Store>.Records via Delete (tolerating the per-store NotFound sentinel).
+// The DeletedRemoteKeys append is what lets chained operations preserve delete intent
+// across intermediate Merges — persistence backends (JSON file, catalog) drive their
+// remote deletes off DeletedRemoteKeys, so a Merge that only removed records and left
+// DeletedRemoteKeys empty would silently drop the delete on every hop.
+//
+// NotFound on the Delete step is tolerated because the source's staged key may
+// legitimately not be present in the destination's Records (RemoteDelete is allowed
+// to stage deletes for records that exist only in the remote backing store), and
+// because we want repeated Merges of the same source to be idempotent.
+//
+// Precedence: a live record in other.<Store>.Records overrides a staged delete in
+// s.<Store>.DeletedRemoteKeys for the same key. This is a side-effect of Upsert,
+// which clears the key from DeletedRemoteKeys whenever a record is upserted (so a
+// direct `RemoteDelete(k); Upsert(rec-with-k)` cancels the staged delete on the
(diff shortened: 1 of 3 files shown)


datastore/memory_datastore.go
```diff
@@ -62,7 +62,28 @@ func (s *MemoryDataStore) WriteMetadata(bundle MetadataBundle, opts ...WriteMeta
...
-// Merge merges the given mutable data store into the current MemoryDataStore.
+// Merge applies records and staged deletions from other onto this MemoryDataStore.
+//
+// Staged deletions (other.<Store>.DeletedRemoteKeys) are propagated by appending the
+// key to s.<Store>.DeletedRemoteKeys via RemoteDelete and then removing the record
+// from s.<Store>.Records via Delete (tolerating the per-store NotFound sentinel).
+// The DeletedRemoteKeys append is what lets chained operations preserve delete intent
+// across intermediate Merges — persistence backends (JSON file, catalog) drive their
+// remote deletes off DeletedRemoteKeys, so a Merge that only removed records and left
+// DeletedRemoteKeys empty would silently drop the delete on every hop.
```
graham-chainlink: do we need to include the reason for this bug fix here?

datastore/memory_datastore.go
```diff
@@ -62,7 +62,28 @@ func (s *MemoryDataStore) WriteMetadata(bundle MetadataBundle, opts ...WriteMeta
...
+// to stage deletes for records that exist only in the remote backing store), and
+// because we want repeated Merges of the same source to be idempotent.
+//
+// Precedence: a live record in other.<Store>.Records overrides a staged delete in
+// s.<Store>.DeletedRemoteKeys for the same key. This is a side-effect of Upsert,
+// which clears the key from DeletedRemoteKeys whenever a record is upserted (so a
+// direct `RemoteDelete(k); Upsert(rec-with-k)` cancels the staged delete on the
+// same store). Staged deletes are only "sticky" on the source side. Callers that
+// need a destination-side staged delete to survive Merge must ensure the source
+// either omits the live record or stages the delete itself.
```
graham-chainlink, line 84: similar here too? Do users need to know all this before using it? I find readin this kind of overwhelming :D

graham-chainlink (dismissed):

ajaskolski (dismissed):

datastore/memory_datastore.go
giogam: no, we can remove that.

datastore/memory_datastore.go
giogam: I think this context is important for explaining the behavior of Merge(), as some aspects of its implementation can otherwise appear opaque or difficult to understand.

ajaskolski (approved):

PR #1047, opened 2026-06-16
feat(evm): add gas boost retry option for contract deploys [CLD-2779]

graham-chainlink:
## Why
In order to take advantage of the contract.NewDeploy method used by the operations-gen cli, we need to add support for gas limit bumping as we currently support for legacy deploy MCMS contract changesets.

Note this already exist in cld-changesets [here](https://github.com/smartcontractkit/cld-changesets/blob/f04fe26c52da79b60e9633d3d06e17f65ecb7893/pkg/family/evm/operations/utils.go#L433), moving and refactoring this to CLDF makes more sense to other users can take advantage of this option.

## Summary

- Adds `RetryDeployWithGasBoost` and `GasBoostConfig` for operations2 EVM contract deploys, increasing gas limit and price on retries after gas-related failures.
- Extends `DeployInput` with optional `GasLimit` and `GasPrice` fields and applies them via `deployTransactOpts` on EVM deploy paths.
- Classifies retryable gas errors using geth sentinels (`errors.Is`) with a string fallback for RPC-wrapped messages.

JIRA: https://smartcontract-it.atlassian.net/browse/CLD-2779


chain/evm/operations2/contract/gas_boost.go
```diff
@@ -0,0 +1,169 @@
...
+// gasRetryableErrorPatterns covers RPC and wrapped errors that no longer carry geth sentinels.
+var gasRetryableErrorPatterns = []string{
+	"out of gas",
+	"gas required exceeds allowance",
+	"intrinsic gas too low",
+	"underpriced",
+	"replacement transaction underpriced",
+	"max fee per gas less than block base fee",
+	"exceeds block gas limit",
+}
```
ecPablo, line 148: nit(noaction): I wonder if we can find better ways to detect retryable errors. I've seen this pattern in gauntlet too, but I've never been happy about it. Wondering if there's better mechanisms nowadays

ecPablo (approved):

chain/evm/operations2/contract/gas_boost.go
graham-chainlink: yeah i know! I dont like this too, currently i am just using a hybrid approach of hardcoded strings [plus sentinel errors](https://github.com/smartcontractkit/chainlink-deployments-framework/pull/1047/changes#diff-af71eeefa7f5385295beff1e5ae569e8849d4b73a1123af82f3da6bfe2104c0aR117-R126) that we know exist

PR #1049, opened 2026-06-16
feat(operations-gen): add zkSync bytecode support [CLD-2780]

graham-chainlink:
## Summary

In order to use operations-gen cli to generate the deploy operations for MCMS contracts , we need to support zksync.

Adds zkSync VM deploy bytecode support to `operations-gen` so generated EVM `Deploy` operations can include both EVM and zkSync bytecode.

Config: 
```yaml
version: "1.0.0"
chain_family: evm

input:
  gobindings_package: "github.com/smartcontractkit/chainlink-deployments-framework/tools/operations-gen/testdata/evm/gobindings"
  zksync_bindings_package: "github.com/smartcontractkit/chainlink-deployments-framework/tools/operations-gen/testdata/evm/zksync_bindings"

output:
  base_path: "."

contracts:
  - contract_name: LinkToken
    version: "1.0.0"
    zksync_bytecode: LinkTokenZkBytecode
    functions:
      - name: transfer
        access: public
      - name: balanceOf
        access: public
      - name: approve
        access: public
```

Closes [CLD-2780](https://smartcontract-it.atlassian.net/browse/CLD-2780).


tools/operations-gen/internal/families/evm/contract.go
```diff
@@ -25,16 +25,18 @@ const (
...
-	OmitDeploy        bool
-	Constructor       *FunctionInfo
-	Functions         map[string]*FunctionInfo
-	FunctionOrder     []string
-	StructDefs        map[string]*structDef
+	Name                  string
+	Version               string
+	PackageName           string
+	GobindingsPackage     string
+	ZkSyncBytecodePackage string
```
ajaskolski: do want to just adds this directly here? maybe some higher lvl 
...ZkSync            *ZkSyncContractInfo

type ZkSyncContractInfo struct {
	BytecodePackage string
	BytecodeSymbol  string
}

or introduce some runtime target ? 

It's not that important when we have only zksync but worth to think how to sort it if we add more l2s

ajaskolski (dismissed):
one small comment but overall looks good to me

ecPablo (dismissed):

tools/operations-gen/internal/families/evm/contract.go
graham-chainlink: done !

ajaskolski (approved):

PR #1072, opened 2026-06-24
Add MCMS to Aptos chain

RodrigoAD:
Add MCMS address to the chain so it can be injected in some adapters from the Datastore

### DIFF
--- a/chain/aptos/aptos_chain.go
+++ b/chain/aptos/aptos_chain.go
@@ -18,6 +18,10 @@ type Chain struct {
 	URL            string
 
 	Confirm func(txHash string, opts ...any) error
+
+	// MCMSAddress is the CCIP MCMS contract on this chain. Zero when unset.
+	// Populated by Aptos deployment/lane changesets from the datastore.
+	MCMSAddress aptoslib.AccountAddress
 }
 
 // Author note: Have to implement the blockhain interface methods explicitly below
--- a/.changeset/public-words-obey.md
+++ b/.changeset/public-words-obey.md
@@ -1,0 +1,5 @@
+---
+"chainlink-deployments-framework": minor
+---
+
+Add MCMS to Aptos chain


chain/aptos/aptos_chain.go
```diff
@@ -18,6 +18,10 @@ type Chain struct {
 	URL            string
 
 	Confirm func(txHash string, opts ...any) error
+
+	// MCMSAddress is the CCIP MCMS contract on this chain. Zero when unset.
```
graham-chainlink, line 22: just checking if this is something specific to CCIP? So we can try avoid anything chain specific here

graham-chainlink:
> Add MCMS address to the chain so it can be injected in some adapters from the Datastore

can you elaborate more on this just so i can understand a bit more? Thanks!

chain/aptos/aptos_chain.go
RodrigoAD: It's CCIP specific. This is to get access to the MCMS address for this [adapter](https://github.com/smartcontractkit/chainlink-ccip/blob/main/deployment/lanes/product.go#L21). MCMS is only needed here for Aptos. I went with this approach as seemed the one with less impact. LMK if you see a better alternative

graham-chainlink: Thanks Rodrigo, i am not too familiar with the lane adapter, is it possible to look up the mcms address using datastore instead? I do feel like having this field in the chain struct is not the most suitable place.

RodrigoAD: > I do feel like having this field in the chain struct is not the most suitable place.

Yeah, I agree. Let me see if we can change the LaneAdapter interface to include MCMS somewhere. 

> is it possible to look up the mcms address using datastore instead? 

We'll actually use Datastore to hydrate the Chain struct in the Changeset Apply. The adapter itself constructs the changeset config, that's why we can't pass custom inputs into the CS

RodrigoAD:
@graham-chainlink We found a way around this. Sorry for the troubles, and thanks for pushing back on it 👍

graham-chainlink:
> @graham-chainlink We found a way around this. Sorry for the troubles, and thanks for pushing back on it 👍

@RodrigoAD awesome news!

PR #1088, opened 2026-07-06
add timelock delay proposal analyzer

RensR:
Adds an always-on analyzer for timelock delays to the proposal analyzer.

### DIFF
--- a/engine/cld/commands/mcms/cmd_analyze_proposal_v2.go
+++ b/engine/cld/commands/mcms/cmd_analyze_proposal_v2.go
@@ -15,6 +15,7 @@ import (
 	"github.com/smartcontractkit/chainlink-deployments-framework/engine/cld/commands/text"
 	"github.com/smartcontractkit/chainlink-deployments-framework/engine/cld/mcms/proposalanalysis"
 	proposalanalysisanalyzer "github.com/smartcontractkit/chainlink-deployments-framework/engine/cld/mcms/proposalanalysis/analyzer"
+	"github.com/smartcontractkit/chainlink-deployments-framework/engine/cld/mcms/proposalanalysis/analyzer/builtin"
 	analysisdecoder "github.com/smartcontractkit/chainlink-deployments-framework/engine/cld/mcms/proposalanalysis/decoder"
 	analysisrenderer "github.com/smartcontractkit/chainlink-deployments-framework/engine/cld/mcms/proposalanalysis/renderer"
 )
@@ -158,7 +159,10 @@ func runAnalyzeProposalV2(cmd *cobra.Command, cfg Config, f analyzeProposalV2Fla
 }
 
 func registerProposalAnalyzers(engine proposalanalysis.AnalyzerEngine, analyzers []proposalanalysisanalyzer.BaseAnalyzer) error {
-	for _, analyzer := range analyzers {
+	defaultAnalyzers := []proposalanalysisanalyzer.BaseAnalyzer{
+		builtin.TimelockDelayValidator{},
+	}
+	for _, analyzer := range append(defaultAnalyzers, analyzers...) {
 		if err := engine.RegisterAnalyzer(analyzer); err != nil {
 			return fmt.Errorf("register proposal analyzer %q: %w", analyzer.ID(), err)
 		}
--- a/engine/cld/mcms/proposalanalysis/analyzer/analyzer.go
+++ b/engine/cld/mcms/proposalanalysis/analyzer/analyzer.go
@@ -24,6 +24,8 @@ type ExecutionContext interface {
 	TimelockAddress(chainSelector uint64) (string, bool)
 	// TimelockAddresses returns all timelock addresses keyed by chain selector.
 	TimelockAddresses() map[uint64]string
+	// ChainMetadata returns proposal chain metadata for building chain-family MCMS clients.
+	ChainMetadata(chainSelector uint64) (mcmstypes.ChainMetadata, bool)
 }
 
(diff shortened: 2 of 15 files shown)


engine/cld/mcms/proposalanalysis/analyzer/builtin/timelock_delay_validator.go
```diff
@@ -0,0 +1,247 @@
...
+	return out
+}
+
+func timelockAddressForChain(execCtx analyzer.ExecutionContext, chainSelector uint64) string {
+	addr, _ := execCtx.TimelockAddress(chainSelector)
+
+	return addr
+}
+
+func readTimelockMinDelay(
```
carte7000, line 188: nit: Not sure if MCMS has this but would be great if we had a registry that do this switching in MCMS so we could get the inspector directly

carte7000 (dismissed):

gustavogama-cll (commented):
@ChrisAmora @graham-chainlink ,
Will defer this to you as you have way more context about the analyzer design but, here are a few initial thoughts:

1. we have a (very old) ticket to implement this validation in the CLDF engine, right after changeset.Apply (that is, *before* creating the proposal). I'm not convinced this type of check should be done so late in the pipeline.
2. If we decide we _do_ want to keep this check as an analyzer, do we want to add a default set of built-ins? Can you envision any other similar built-in analyzers? Would it make sense to incorporate Loren's risk analyzer under the same model?
3. the timelock_inspectors.so implementation should definitely be moved to the mcms library `chainwrappers` package

graham-chainlink:
> > 1. we have a (very old) ticket to implement this validation in the CLDF engine, right after changeset.Apply (that is, _before_ creating the proposal). I'm not convinced this type of check should be done so late in the pipeline.

@gustavogama-cll 

I do kind of agree, ideally the earlier the better, mainly i think that the validation here is something we are able to do earlier as we already have the data, eg we dont have to wait for the proposal to be generated in order to check the min delay values. But i guess this is something we have to do on our side (CLD) , however this analyzer is better than nothing if on the CLD side we are unable to add this validation in the short term.

engine/cld/mcms/proposalutils/timelock_inspectors.go
```diff
@@ -0,0 +1,120 @@
...
+	blockChains chain.BlockChains,
+	chainSelector uint64,
+	metadata mcmstypes.ChainMetadata,
+) (mcmssdk.TimelockInspector, error) {
+	chainAccessor := cldfmcmsadapters.Wrap(blockChains)
+
+	return buildTimelockInspector(&chainAccessor, mcmstypes.ChainSelector(chainSelector), metadata)
+}
+
+func buildTimelockInspector(
```
gustavogama-cll, line 38: this is now available in [mcms v0.51.0](https://github.com/smartcontractkit/mcms/blob/main/chainwrappers/timelock_inspectors.go#L39)

engine/cld/mcms/proposalanalysis/analyzer/analyzer.go
```diff
@@ -14,6 +16,16 @@ type ExecutionContext interface {
...
+	// ProposalAction returns the timelock action for the proposal under analysis.
+	ProposalAction() mcmstypes.TimelockAction
+	// ProposalDelay returns the proposal-level timelock delay.
+	ProposalDelay() mcmstypes.Duration
+	// TimelockAddress returns the timelock contract that executes operations on the chain.
+	TimelockAddress(chainSelector uint64) (string, bool)
+	// TimelockAddresses returns all timelock addresses keyed by chain selector.
+	TimelockAddresses() map[uint64]string
+	// ChainMetadata returns proposal chain metadata for building chain-family MCMS clients.
+	ChainMetadata(chainSelector uint64) (mcmstypes.ChainMetadata, bool)
```
graham-chainlink, line 28: I am updating the API so we get access to these in the renderer instead of having to do this, will have a PR up for this.

https://github.com/smartcontractkit/chainlink-deployments-framework/pull/1106

engine/cld/commands/mcms/cmd_analyze_proposal_v2.go
```diff
@@ -158,7 +159,10 @@ func runAnalyzeProposalV2(cmd *cobra.Command, cfg Config, f analyzeProposalV2Fla
 }
 
 func registerProposalAnalyzers(engine proposalanalysis.AnalyzerEngine, analyzers []proposalanalysisanalyzer.BaseAnalyzer) error {
-	for _, analyzer := range analyzers {
+	defaultAnalyzers := []proposalanalysisanalyzer.BaseAnalyzer{
+		builtin.TimelockDelayValidator{},
+	}
+	for _, analyzer := range append(defaultAnalyzers, analyzers...) {
```
graham-chainlink, line 165: i wonder if instead of registering here, we register the built in analyzer deeper in the framework so we treat this as a proper built in, currently this is registered on the cli command layer.

I was thinking somewhere [around here](https://github.com/smartcontractkit/chainlink-deployments-framework/blob/7aed8eeb37a1b64069d4d5f81aa42d36aa34e43f/engine/cld/mcms/proposalanalysis/engine.go#L56-L73) where we create the analyzer engine, what do you think @ChrisAmora ?

graham-chainlink:
Hi @RensR  we are making a few changes (Gustavo and me per the comments in this PR) to make the change in this PR a bit easier.

RensR:
We'll resolve the delay instead

graham-chainlink:
For reference, this PR is superseeded by https://github.com/smartcontractkit/chainlink-deployments-framework/pull/1114

PR #1090, opened 2026-07-06
feat: ed25519

ChrisAmora:
This pull request introduces support for managing Ed25519 keys in AWS KMS for non-EVM chains. It adds configuration structures, environment variable bindings, and a new implementation for signing and public key retrieval using Ed25519 keys in KMS.


chain/internal/kms/ed25519.go
```diff
@@ -0,0 +1,121 @@
...
+// an Ed25519 key.
+type Ed25519Client interface {
+	GetPublicKey(ctx context.Context, in *kmsv2.GetPublicKeyInput, optFns ...func(*kmsv2.Options)) (*kmsv2.GetPublicKeyOutput, error)
+	Sign(ctx context.Context, in *kmsv2.SignInput, optFns ...func(*kmsv2.Options)) (*kmsv2.SignOutput, error)
+}
+
+var _ Ed25519Client = (*kmsv2.Client)(nil)
+
+// NewEd25519Client constructs an aws-sdk-go-v2 KMS client for the given region.
+func NewEd25519Client(ctx context.Context, keyRegion string) (Ed25519Client, error) {
```
gustavogama-cll: shouldn't we return `*kmsv2.Client`? ([ref](https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b))

chain/internal/kms/ed25519_test.go
gustavogama-cll, line 1: nit: most of CLDF uses table-driven unit tests

engine/cld/config/env/config.go
```diff
@@ -18,6 +18,15 @@ type KMSConfig struct {
 	KeyRegion string `mapstructure:"key_region" yaml:"key_region"` // Secret: AWS KMS Key Region (e.g. us-west-1)
 }
 
+// KMSEd25519Config is the configuration for the Ed25519 AWS KMS key used by non-EVM chains.
+//
+// WARNING: This data type contains sensitive fields and should not be logged or set in file
+// configuration.
+type KMSEd25519Config struct {
```
gustavogama-cll: have you considered adding a "scheme" (or "type", or "algorithm") attribute to `KMSConfig` instead?

gustavogama-cll (commented):
looks good to me but I'd want to hear what the others think about the way we're exposing the configuration option. I personally think it would be better to add an optional "scheme" attribute, rather than defining a top level `kms_ed25519` key.

chain/internal/kms/ed25519.go
ChrisAmora: good catch

engine/cld/config/env/config.go
ChrisAmora: Yeah, but that would allow only one curve per run A collection keyed by scheme could work but breaks the current schema

gustavogama-cll: and making it an array is an issue because we don't want to change the current parser? It's yaml after all...

graham-chainlink: unrelated to scheme, you can also do 
```go
type KMSEd25519Config = KMSConfig // type alias, not a new struct
```
Since both struct are the same if we go down this path,

however i do like the scheme idea but then we can only pick EVM or non EVM in the config.

Ideally if we can redesign from scratch, it could be something like 
```go
type KMSKeysConfig struct {
    Secp256k1 KMSKeyConfig
    Ed25519   KMSKeyConfig
}
```

but i think so far we will only have evm and non evm, what are the odds of having another type of KMS key?

gustavogama-cll (approved):

PR #1094, opened 2026-07-07
Add a error based retry retry_policy

carte7000:
Add a error based retry retry policy

1. No policy should presever the current behaviour
2. Delay can be specified by matching an error
3. Add nonce too low and no code at error in the default evm chain loader

### DIFF
--- a/engine/cld/chains/chains.go
+++ b/engine/cld/chains/chains.go
@@ -528,8 +528,8 @@ func (l *chainLoaderEVM) Load(ctx context.Context, selector uint64) (fchain.Bloc
 				DialTimeout:        2 * time.Second,
 				HealthCheckTimeout: 15 * time.Second, // high concurrency needs more headroom than the 2s default
 				ErrorPolicies: []evmclient.ErrorRetryPolicy{
-					nonceTooLowRetryPolicy(2 * time.Second),
-					noContractCodeRetryPolicy(2 * time.Second),
+					nonceTooLowRetryPolicy(5 * time.Second),
+					noContractCodeRetryPolicy(5 * time.Second),
 				},
 			}
 		},
--- a/engine/cld/chains/chains.go
+++ b/engine/cld/chains/chains.go
@@ -527,6 +527,10 @@ func (l *chainLoaderEVM) Load(ctx context.Context, selector uint64) (fchain.Bloc
 				DialDelay:          10 * time.Millisecond,
 				DialTimeout:        2 * time.Second,
 				HealthCheckTimeout: 15 * time.Second, // high concurrency needs more headroom than the 2s default
+				ErrorPolicies: []evmclient.ErrorRetryPolicy{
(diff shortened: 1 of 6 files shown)


engine/cld/chains/chains.go
```diff
@@ -527,6 +527,10 @@ func (l *chainLoaderEVM) Load(ctx context.Context, selector uint64) (fchain.Bloc
 				DialDelay:          10 * time.Millisecond,
 				DialTimeout:        2 * time.Second,
 				HealthCheckTimeout: 15 * time.Second, // high concurrency needs more headroom than the 2s default
+				ErrorPolicies: []evmclient.ErrorRetryPolicy{
+					nonceTooLowRetryPolicy(2 * time.Second),
```
RensR: Given these are now per error, maybe we can even go to 5s

engine/cld/chains/evm_retry_policies.go
```diff
@@ -0,0 +1,39 @@
...
+// isNoContractCodeError reports whether err indicates contract code is not yet available.
+func isNoContractCodeError(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := strings.ToLower(err.Error())
+
+	return strings.Contains(msg, "no code at") ||
+		strings.Contains(msg, "no contract code") ||
+		strings.Contains(msg, "empty string")
```
RensR: Hmm this feels not super scopes. Is this one really needed? There's a million empty string errors

RensR (approved):

engine/cld/chains/chains.go
```diff
@@ -527,6 +527,10 @@ func (l *chainLoaderEVM) Load(ctx context.Context, selector uint64) (fchain.Bloc
 				DialDelay:          10 * time.Millisecond,
 				DialTimeout:        2 * time.Second,
 				HealthCheckTimeout: 15 * time.Second, // high concurrency needs more headroom than the 2s default
+				ErrorPolicies: []evmclient.ErrorRetryPolicy{
+					nonceTooLowRetryPolicy(5 * time.Second),
+					noContractCodeRetryPolicy(5 * time.Second),
```
ecPablo, line 532: Are we sure we can blanket add these to all evm chains across all CLL products without unintended consequences? I'm trying to think of scenarios where we might want different error policies per domain or per environment. Maybe this is something we might want to configure more granularly?

carte7000: I think the worst that can happen here is that the pipeline is slower to execute in legitimate case. Are you seeing some cases where it cause a correctness issue?

RensR: I wouldn't call this blocking. Given the narrow scope of when it triggers, this is purely an upside for all CLL products. These errors are RPC related, not product. Other products use the same RPCs

ecPablo: Sure I can't thing of any other issues myself from this either so let's go with it

ecPablo (approved):

PR #1112, opened 2026-07-09
feat(proposalanalysis): builtin analyzer infrastructure

graham-chainlink:
Introduce a framework for supporting builtin analyzers so it is more scalable when we add more builtin analyzers, currently adding new builtin analyzers require too many touch point such as the template , constants etc.

This framework allow the idea of a generic template for any built in analyzers , so we dont have to keep updating the master template , and instead just define a new template file for the new analyzers and register them

[This is an example](https://github.com/smartcontractkit/chainlink-deployments-framework/pull/1113) of how a new built-in analyser using this framework can be added


engine/cld/mcms/proposalanalysis/analyzer/builtin/register.go
```diff
@@ -0,0 +1,10 @@
+package builtin
+
+import (
+	"github.com/smartcontractkit/chainlink-deployments-framework/engine/cld/mcms/proposalanalysis/analyzer"
+)
+
+// RegisterAll registers built-in proposal analyzers with the registry.
+func RegisterAll(_ *analyzer.Registry) {
+	// to be added in new PR
```
graham-chainlink, line 9: New builtin analysers will be registered here using `register(timelockdelay.Validator{})`

Preparation for https://github.com/smartcontractkit/chainlink-deployments-framework/pull/1088

engine/cld/mcms/proposalanalysis/renderer/templates/markdown/proposal.tmpl
```diff
@@ -6,6 +6,9 @@
 
 {{ end -}}
 
+{{- /* Built-in analyzer reports (pre-rendered) */ -}}
+{{- if .BuiltinSections }}{{ .BuiltinSections }}{{ end }}
```
graham-chainlink, line 10: All built-in analyzers can use this same section

PR #1114, opened 2026-07-10
feat(pipeline): correct timelock delay from on-chain minDelay during run

graham-chainlink:
## Summary

Correct schedule proposal timelock delays during durable-pipeline `run`, immediately after changeset Apply and before proposals are persisted.

- Add `engine/cld/mcms/timelockdelay` to read on-chain `minDelay` per timelock chain (parallel RPC reads) and bump unset/too-low delays to the max across chains
- Relax `MCMSTimelockProposalInput.Validate()` to allow zero delay for schedule actions; durable-pipeline resolves it when every timelock chain is readable
- Fail the pipeline when delay is unset and on-chain verification fails; warn and continue when the author set an explicit delay but RPC verification fails

Closes [CLD-3058](https://smartcontract-it.atlassian.net/browse/CLD-3058).

## Timelock delay correction behaviour

Applies only to **schedule** proposals (`TimelockActionSchedule`). Non-schedule actions are skipped.

| Proposal delay | On-chain verification | Pipeline | Resulting delay |
|---|---|---|---|
| Unset (≤0) | All chains OK, `max(minDelay) > 0` | Continue | Set to `max(minDelay)` across timelock chains |
| Unset (≤0) | Any RPC/read failure | **Fail** (`ErrUnsetTimelockDelayUnverified`) | Unchanged |
| Unset (≤0) | All OK but `max(minDelay) = 0` | **Fail** (`ErrUnsetTimelockDelayUnverified`) | Unchanged |
| Unset (≤0) | No timelock addresses in proposal | **Fail** (`ErrUnsetTimelockDelayUnverified`) | Unchanged |
| Explicit, below `max(minDelay)` | All chains OK, `max(minDelay) > 0` | Continue | Set to `max(minDelay)` |
| Explicit, ≥ `max(minDelay)` | All chains OK | Continue | Unchanged |
| Explicit (any value) | Any RPC/read failure | Continue (warn) | Unchanged |
| Explicit (any value) | No timelock addresses | Continue (warn) | Unchanged |
| Explicit (any value) | All OK but `max(minDelay) = 0` | Continue | Unchanged |

**Rationale:** An unset delay cannot be safely inferred without on-chain verification, so any verification failure is a hard error. An explicit delay was author-chosen; if RPC fails we warn and leave it unchanged rather than block the pipeline.

### DIFF
--- a/engine/cld/mcms/timelockdelay/correct.go
+++ b/engine/cld/mcms/timelockdelay/correct.go
@@ -1,0 +1,126 @@
+package timelockdelay
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/smartcontractkit/mcms"
+	mcmstypes "github.com/smartcontractkit/mcms/types"
+
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain"
+	"github.com/smartcontractkit/chainlink-deployments-framework/pkg/logger"
+)
+
+// MinDelayLookup reads on-chain minDelay for one timelock chain entry.
+type MinDelayLookup func(
+	ctx context.Context,
+	proposal *mcms.TimelockProposal,
+	chainSelector uint64,
+	timelockAddress string,
+) (mcmstypes.Duration, error)
+
+// CorrectTimelockDelays updates schedule proposal delays using on-chain minDelay from blockChains.
+// Unset or too-low delays are bumped to the max on-chain minDelay across timelock chains when
+// minDelay can be read for every timelock chain. An unset delay fails the call if any chain
+// cannot be read or on-chain minDelay is zero. An explicitly set delay is left unchanged when
+// on-chain minDelay cannot be verified.
+func CorrectTimelockDelays(
+	ctx context.Context,
+	lggr logger.Logger,
+	blockChains chain.BlockChains,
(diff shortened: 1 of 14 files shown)


engine/cld/mcms/timelockdelay/correct.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil
+	}
+
+	unsetDelay := proposal.Delay.Duration <= 0
+
+	if len(proposal.TimelockAddresses) == 0 {
+		if unsetDelay {
+			return fmt.Errorf("%w: no timelock addresses in proposal", ErrUnsetTimelockDelayUnverified)
+		}
+		lggr.Warnw("skipping timelock delay correction: no timelock addresses in proposal",
```
RensR, line 78: Is this even a valid case?

RensR (approved):

KodeyThomas (approved):

ajaskolski (approved):

engine/cld/mcms/timelockdelay/correct.go
graham-chainlink: probably more of a just in case as it can happen due to a bug somewhere.

ChrisAmora (approved):

PR #1136, opened 2026-07-28
skip inspector building if it needs additional fields

FelixFan1992:


gustavogama-cll:
sorry @FelixFan1992 but I don't think this is the direction we should head. Let us align on the path forward on the slack thread, but I believe you have been able to execute the changeset without modifying CLDF.

